PR #3, opened 2018-03-01
make path to database configurable

sascha-andres:
By setting an environment variable to a path to the database you can store your db where ever you want

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -2,14 +2,15 @@ package cmd
 
 import (
 	"fmt"
-	"github.com/RadhiFadlillah/shiori/database"
 	"github.com/spf13/cobra"
 	"os"
+	db "github.com/RadhiFadlillah/shiori/database"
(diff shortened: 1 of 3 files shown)


cmd/root.go
```diff
@@ -25,3 +26,14 @@ func Execute() {
 		os.Exit(1)
 	}
 }
+
+func init () {
```
RadhiFadlillah: Hi @sascha-andres, thanks for the PR.

It looks good, however I think opening database should be done in `main.go` instead here. The reasons are :

- When opening database, there is a possibility that panic happened. I personally think `init` must never panic, because `init` always executed, even if the package only imported and not used.
- For better package separation. As you can see, the [`Database`](https://github.com/RadhiFadlillah/shiori/blob/master/database/database.go#L10) is declared as interface because I have plan to support other DBMS in the future. The `cmd` is supposed to only use the `Database`, without caring what kind of database it is. That's why I think opening database should be done in `main.go`.

PR #11, opened 2018-03-01
Add dockerfile

viktorstrate:
Add [Docker](https://www.docker.com/) support. This file will download dependencies and install shiori inside an image. This is extremely useful for server maintainers, because it's secure and easy.

Run `docker build -t shiori .` to build the image
Run `docker run -p 8080:8080 -d --rm --name shiori shiori` to start a docker container
Run `docker exec shiori shiori <COMMAND>` to interact with the container

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM golang:1.8
+
+WORKDIR /go/src/shiori
+COPY . .
+
+RUN go get -d -v ./...
+RUN go install -v ./...
+
+CMD ["shiori", "serve"]


maxmeyer:
#27 and #11 trying to solve the same problem.

maxmeyer:
As written in #27: I suggest to use multi stage builds instead: https://docs.docker.com/develop/develop-images/multistage-build/#name-your-build-stages. There's no need for a full "go"-aware alpine image for that. The link contains a suggestion about how to build go projects with that docker feature.

viktorstrate:
I don't know much about Go. And I made a dockerfile for my own needs, and thought I would share it with you. But I do not know enough about Go to improve it.

Please see #27 for a better implementation

PR #21, opened 2018-03-02
Cleaned up command imports and descriptions

muesli:
Removed the trailing dots which are a bit uncommon for descriptions and cleaned up the import sorting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,17 +39,17 @@ Usage:
   shiori [command]
 
 Available Commands:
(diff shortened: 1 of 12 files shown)


RadhiFadlillah:
Hi, this looks nice. I'm used to use `gofmt` which doesn't group the import path. I guess I will switch to `goimports` from now on.

muesli:
@RadhiFadlillah Absolutely recommended indeed!

PR #27, opened 2018-03-02
Dockerfile + README.md usage

dexafree:
This PR adds a Dockerfile that will build Shiori based on the official Go alpine image (by using the alpine version, the resulting image will be smaller).

Comparison between using the standard image (left) vs the Alpine one (right):

Comparison

It also adds a little "Usage with Docker" section to the Readme file that will help the users self-host their own shiori containers, with the ability to persist the database across container recreations.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,4 +13,6 @@ RUN apk --no-cache add git build-base \
 && go install -v ./... \
 && touch shiori.db
 
+EXPOSE 8080
+
 CMD ["shiori", "serve"]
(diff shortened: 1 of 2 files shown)


maxmeyer:
I suggest to use multi stage builds instead: https://docs.docker.com/develop/develop-images/multistage-build/#name-your-build-stages. There's no need for a full "go"-aware alpine image for that. The link contains a suggestion about how to build go projects with that docker feature.

maxmeyer:
#11 and #27 trying to solve the same problem.

RadhiFadlillah:
Closed since #66 has been merged.

PR #37, opened 2018-03-03
Import tags from bookmark.

### LINKED ISSUE #34: Tags are not imported from NETSCAPE-Bookmark-file
I have NETSCAPE-Bookmark-file exported from Firefox. It contains `TAGS` parameter with tags for each bookmark. It would be nice if shiori could import this tags.

skx:
This update allows reading tags from the bookmark, at import-time.  This will close #34.


skx:
Fixed in :+1: 

https://github.com/RadhiFadlillah/shiori/commit/837c4701760dc479be7e0b4b7dabe2f1408b1be0

PR #38, opened 2018-03-03
add travis ci

ddulic:
Add the option to deploy releases via Travis CI deployments for linux, osx and windows binaries.

If you haven't used travis before just create an account on https://travis-ci.org/ connect and enable this github repo. Should be self explanatory.
One thing you will have to do is add your `$TOKEN` via travis env variables. You have multiple ways of doing this which you can read [here](https://docs.travis-ci.com/user/environment-variables/)
I used [Defining Variables in Repository Settings](https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings) in my example.

Once you have configured everything simply tag releases as you did before and travis will create a release on every tagged commit :)

Hopefully this will be of some use :+1:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,39 @@
+language: go
+sudo: false
+matrix:
+  include:
+    - go: 1.x
+      env: LATEST=true
+    - go: 1.9
+    - go: 1.8
+
+before_install:
+  - go get github.com/mitchellh/gox # go tool for cross compiling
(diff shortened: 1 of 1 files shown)


RadhiFadlillah:
Hey @ddulic, thanks for the PR. 

I've used Travis before, but only for simple testing and not for deploying releases, so I'm really thankful that you go and make this.

PR #42, opened 2018-03-04
Change module path to relative, add check for empty tag on import

generall:


RadhiFadlillah:
Hi @generall, thanks for the PR.

I've accepted the PR, but I've reverted the relative paths back to absolute paths. There has been a [lot](https://github.com/golang/go/issues/3515#issuecomment-66066361) [of](https://groups.google.com/d/topic/golang-nuts/n9d8RzVnadk/discussion) [discussions](https://groups.google.com/d/topic/golang-nuts/zFVF7W-Y_y0/discussion) about absolute vs relative import paths, and the conclusion is the *relative import paths is bad*. The reasons are :

- You can't `go install` and `go get` packages with relative import.
- While it might make directories relocatable, it makes source files and packages fixed with respect to their surroundings.

generall:
It was unclear to me how to compile package with my local modifications and without changing path to relative since I have no access to your github repository.

iwanbk:
You could clone your repo to `$GOPATH/github.com/RadhiFadlillah/shiori` instead of `$GOPATH/github.com/generall/shiori`

PR #46, opened 2018-03-06
Prepend environment path to database path

AbeEstrada:
This fixes an error when the `ENV_SHIORI_DB` is defined, it doesn't open a database file because the database file name isn't on the `databasePath` variable.

### DIFF
--- a/main.go
+++ b/main.go
@@ -3,6 +3,7 @@ package main
 
 import (
 	"os"
+	fp "path/filepath"
 
 	"github.com/RadhiFadlillah/shiori/cmd"
 	db "github.com/RadhiFadlillah/shiori/database"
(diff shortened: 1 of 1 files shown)


sascha-andres:
Except when you set the complete database (eg `/home/user/.shiori/bookmarks.db`) as I do. Then you would end up with `/home/user/.shiori/bookmarks.db/shiori.db`, a path that probably does not exist and shall not exist.

IMO a better approach would be to check whether `ENV_SHIORI_DB` is a directory to append `/shiori.db` else use the file. In each case, if the directory does not exist, error out.

RadhiFadlillah:
@AbeEstrada thank you very much and sorry for late response.

@sascha-andres I've added code to check if environment variable points to directory in commit [c79b117](https://github.com/RadhiFadlillah/shiori/pull/46/commits/c79b117ddb5cf0161be7724e1b272ae5ffd6f4fc).

PR #48, opened 2018-03-06
additions for tag cloud

s-frostick:
created a tag cloud and new api endpoint (api/tags) that returns a list of tags and the number of occurrences.


s-frostick:
sorry i think i have made a mistake. will fix

PR #52, opened 2018-03-06
Set server timeouts

### LINKED ISSUE #51: Add server timeouts
Add server timeouts to avoid leaking file descriptors in case of slow or disappearing clients.

peteretelej:
Set server timeouts to avoid leaked descriptors on slow/disappearing clients or slow http attacks.

Fixes #51


RadhiFadlillah:
Thank you very much

Issue #29, opened 2018-03-02
Feature Request: Import/Seed Shiori with Pocket bookmarks


codeanpeace:
This could be a killer feature to drive adoption of shiori.

RadhiFadlillah:
I’m closing this issue because it has been implemented, and also inactive for almost a month. Feels free to reopen it if needed.

Thank you.

PR #63, opened 2018-03-08
new pocket command 

s-frostick:
#29  Added pocket integration using  [https://github.com/motemen/go-pocket](url).

User needs to set a consumer key or create a new one; https://getpocket.com/developer/apps/new

Then can run the following commands 

`shiori pocket key <APIKEY>`
to set the api key

`shiori pocket login` 

The login command will open a browser where the user can login to pocket to authorize the request.

The accesstoken and consumer key are stored in $HOME/.shiori/config.json

Then can run 
`shiori pocket import`
to actually import items from pocket

by default this will not import items that have been archived in pocket, for this run with

`shiori pocket import --all`

If just want to import items with a particular tag can use

`shiori pocket import --tag <TAG>`

Note: the import command imports tags with spaces from pocket although the current web interface is unable to handle these correctly.

### DIFF
--- a/cmd/pocket.go
+++ b/cmd/pocket.go
@@ -71,8 +71,7 @@ var (
 		Run: func(cmd *cobra.Command, args []string) {
 			offline, _ := cmd.Flags().GetBool("offline")
 			all, _ := cmd.Flags().GetBool("all")
-            tag, _ := cmd.Flags().GetString("tag") 
-
+			tag, _ := cmd.Flags().GetString("tag")
 
 			apikey := viper.Get("pocket.apikey")
 			if apikey == nil {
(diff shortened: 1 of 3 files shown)


RadhiFadlillah:
Hi @s-frostick, thanks for the PR. However, I found this is too complicated for the user, because they need to create Pocket application first. 

To solve this, I've made it so user can just import his Pocket's account by using the [exported HTML file](https://help.getpocket.com/article/1015-exporting-your-pocket-list) in [`a74773b`](https://github.com/RadhiFadlillah/shiori/commit/a74773bd144843822596ff70c731677b378bcad3). The only downside is it won't import the bookmark's title from Pocket. However, since (AFAIK) all title in Pocket is resolved from the webpage's metadata, user can easily do it himself by using `shiori update` command.

PR #66, opened 2018-03-09
Extended docker support

maxmeyer:
# Preface

This incorparates all changes from #27 (/cc @dexafree) rebased to current master and reduces the image size by using the multistaged build feature from docker. This also reduces the surface for attacks by reducing the amount of installed packages.

# Additions

* Multistaged docker build - minimal Alpine Linux Image for the image
* Build/Setup-scripts to make it easier for newcomers to get started
* Extend documentation about how to run the docker container

PR #70, opened 2018-03-10
Improve contributor experience

maxmeyer:
# Preface

This extracts some commits from #66 to focus discussion on related things.

# Reason for separate PR

> BTW, I think we don't really need to add those additional scripts, especially since go command is already simple enough to use. For example, setup is just wrapper for go get and build is just wrapper for go build.

# Changes introduced

* Add scripts
* Documentation for those scripts
* ~~Migrate dependency handling to dep as it is stable and can handle that task well~~ This failed and was removed in a later iteration
* Migrate logic from `.travis.yml` to scripts to make it usable for contributors as well
* Replace logic in `Dockerfile` with scripts

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,3 @@
-bin/
 README.md
 Dockerfile
 LICENSE
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,20 +7,13 @@ matrix:
     - go: 1.9
     - go: 1.8
 
-before_install:
-  - go get github.com/mitchellh/gox # go tool for cross compiling
-  - go get github.com/inconshreveable/mousetrap # needed for windows builds
-
 install:
-  - go get github.com/RadhiFadlillah/shiori
+  - bin/setup
 
 script:
-  - go get -t -v ./...
   - diff -u <(echo -n) <(gofmt -d .)
-  - go vet $(go list ./... | grep -v /vendor/)
-  - go test -v -race ./...
-  # only build binaries from the latest Go release.
-  - if [ "${LATEST}" = "true" ]; then gox -os="linux darwin windows" -arch="amd64" -output "shiori_{{.OS}}_{{.Arch}}" -ldflags "-X main.Rev=`git rev-parse --short HEAD`" -verbose ./...; fi
+  - bin/test
+  - if [ "${LATEST}" == "true" ]; then OSSES="linux darwin windows" bin/build; else bin/build; fi
 
 deploy:
   provider: releases
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,15 +5,15 @@ RUN apk update \
 
 WORKDIR /go/src/github.com/RadhiFadlillah/shiori
 COPY . .
-RUN go get -d -v ./...
-RUN go build -o shiori main.go
+RUN bin/setup
+RUN bin/build
 
 FROM alpine:latest
 
 ENV ENV_SHIORI_DB /srv/shiori.db
 
 RUN apk --no-cache add dumb-init ca-certificates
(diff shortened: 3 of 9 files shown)


maxmeyer:
@RadhiFadlillah 
You're right about that the scripts are quite small. 

Personally I think, that 
* using `go get` and the like is an "implementation detail" for (new) contributors
* you don't need to change more than one place to add new logic to the scripts - no change to documentation and no change of .travis.yml is required

Reasons for the proposa:
I'm involved in quite a few projects to fix things, to add new features, to fix documentation, to dockerize the project etc. In most cases it's cumbersome to get started. There's no documentation what is required to setup the project, what kind of tests need to be run and how I can build the project. That's why I like the proposal found in here https://githubengineering.com/scripts-to-rule-them-all/. Having a simple script is an easy entry point. Giving them clear names makes their use case obvious. Since you've got the script I don't need to care which tool is used to bootstrap the project etc.

maxmeyer:
@RadhiFadlillah Rebased to HEAD.

peteretelej:
@maxmeyer sorry but I think using `go get` keeps with Go convention and IMHO a makefile + scripts would be an unnecessary overhead, since the app's build process is quite straightforward.

At the moment, these things seem straightforward;
- to do fetch source,update, build and install this is the only thing to run: 
```go get -u github.com/RadhiFadlillah/shiori``` 
- building binaries from source is simply `go build`
- testing is `go test ./...`
- to build the docker image `docker build -t myname/shiori .` ~ this may be better than what the build script does (since this is a custom build + can't be pushed to Radhi's docker acc) . 
- to use the _official_ docker container is `docker run -P radhifadlillah/shiori`

>  I'm involved in quite a few projects to fix things, to add new features, to fix documentation, to dockerize the project etc. In most cases it's cumbersome to get started. There's no documentation what is required to setup the project, what kind of tests need to be run and how I can build the project.

Not usu the case with go apps (like explained above), unless am not getting something. Maybe the documentation may need sprucing up to make the commands clear

maxmeyer:
> to do fetch source,update, build and install this is the only thing to run:
> go get -u github.com/RadhiFadlillah/shiori

That's true for users

> building binaries from source is simply go build

Please, have a look at [.travis.yml](https://github.com/RadhiFadlillah/shiori/blob/master/.travis.yml#L23), for platform dependent builds `gox` is involved, so it's not just `go build`. That's not obvious for new contributors - at least not for me.

> testing is go test ./...

Again, please have a look at [.travis.yml](https://github.com/RadhiFadlillah/shiori/blob/master/.travis.yml#L20). `go vet` is also involved, which is again not obvious for first timers.

> to build the docker image docker build -t myname/shiori . ~ this may be better than what the build script does (since this is a custom build + can't be pushed to Radhi's docker acc) .

I'm not sure what you mean with ` can't be phttps://github.com/RadhiFadlillah/shiori/blob/master/.travis.yml#L23ushed to Radhi's docker acc`. Of course you can do `docker push radhifadlillah/shiori`.

> Not usu the case with go apps (like explained above), unless am not getting something. Maybe the documentation may need sprucing up to make the commands clear

For the normal user tasks, this is clear. But today there's a lot of knowledge hidden within the `.travis.yml`-file. 

I don't think it's neccessary to have scripts, but I would argue, that all knowledge to build/test the app should be obvious and either clearly documented or abstracted by a utility.

peteretelej:
please ignore the `.travis.yml` file. `gox` is not part of the build process, the build process is simply `go build`, gox is used to automate the build for various platform binaries (in this case travis uses it for deploying the releases to github). `go vet` is usually included as part of most editor's linters (hence you really won't be using it explicitly, if you do [gometalinter](https://github.com/alecthomas/gometalinter) would be the way to go)

> all knowledge to build/test the app should be obvious and either clearly documented or abstracted by a utility.

True.

PR #87, opened 2018-04-23
Add MySQL database support

### LINKED ISSUE #81: can shiori work with different database?
if i want to use with mysql or postgresql, can shiori work with those db?

peteretelej:
Add support for MySQL as the underlying database.

Closes #81


peteretelej:
@gissehel true, though the issue was if it can work with a different db. The database is an interface so you could probably add any other db type even a key-value store as long as it implements the interface.

DWSR:
Would love to see this as I would definitely run this as a service on Kubernetes.

PR #97, opened 2018-06-06
feature/database name: allow setting a different database name

sascha-andres:
With two environment variables (ENV_SHIORI_DIR and ENV_SHIORI_DB) the storage location can be customized completely


RadhiFadlillah:
Thank you

RadhiFadlillah:
In retrospect, I don't think it's good idea to use the environment variable `ENV_SHIORI_DB` again. The reason is we already using `ENV_SHIORI_DIR` to set where `shiori` will save its data. Therefore, I think there are no advantage for user to be able to set custom database name.

Issue #101, opened 2018-06-23
Ability to specify Listen-Address


contradictioned:
There is no way to specify a listen address (e.g., 127.0.0.1, or a specific public IP if you have multiple at hand). If I understand correctly, it is because the url for the HTTP-Server is set like here https://github.com/RadhiFadlillah/shiori/blob/master/cmd/serve/root.go#L60

Would you accept a PR where I introduce an option `-l, --listen addr` for specifying a listen address?

RadhiFadlillah:
Hi @contradictioned, sorry for late reply.

> Would you accept a PR where I introduce an option `-l, --listen addr` for specifying a listen address?

Yes, feels free to do that, and thank you.

RadhiFadlillah:
Since the `ramadhan` branch is merged, the PR has been merged. However, I change the flag from `-listen` into `-address` because IMHO it's more clear. 

Since it has been implemented, I will close this issue. Feels free to reopen it if needed.

PR #118, opened 2018-10-17
Adds flag for specifying a listen address as suggested in #101.

contradictioned:


aminehmida:
Any chance this can be merged? It will be ideal if I can make shiori listen on a specific interface only. Thanks.

PR #122, opened 2019-01-24
add check command

sascha-andres:
This PR adds a check command that iterates over all bookmarks and tries to call them and prints a status information.

Optionally you can have unsuccessful sites be deleted from the database


cmd/root.go
```diff
@@ -107,6 +115,9 @@ func NewShioriCmd(db dt.Database, dataDir string) *cobra.Command {
 	addCmd.Flags().StringSliceP("tags", "t", []string{}, "Comma-separated tags for this bookmark.")
 	addCmd.Flags().BoolP("offline", "o", false, "Save bookmark without fetching data from internet.")
 
+	checkCmd.Flags().BoolP("index-only", "i", false, "Only print the index of bookmarks")
+	checkCmd.Flags().BoolP("remove", "r", false, "Automatically remove non exiting bookmarks")
```
techknowlogick: `exiting` -> `existing`

sascha-andres: @techknowlogick thanks for pointing this out, pushed a new commit to adress this

RadhiFadlillah:
@sascha-andres sorry for late, late response.

Thanks for the PR, I've never thought to add command specifically to check if bookmarks still available online. Because of the major rewrite, I can't merge this PR. Instead I just copy and modify it, and now it has been implemented in [2eb6415](https://github.com/go-shiori/shiori/commit/2eb6415fa883b8c7268a72064bf14c38e46babe9).

Thanks !

PR #149, opened 2019-09-06
Add LDAP authentication, TOML config and Dockerfile

ynsta:


internal/auth/ldapauth.go
```diff
@@ -0,0 +1,161 @@
...
+
+	dn := sr.Entries[0].DN
+	if loginField != "" {
+		return dn, sr.Entries[0].GetAttributeValue(loginField), nil
+	}
+	return dn, username, nil
+}
+
+// Search connect and search for a username in the ldap, add the entry
+func (la *LDAPAuth) Search(username string, group string, loginField string) (string, string, error) {
```
RadhiFadlillah: I'm not really familiar with how go-ldap works, but does `*ladp.Conn` must be created for every search ?

I never works with LDAP, however in websocket usually the connection is only created once at the beginning. Then all search and query will reuse of that connection.

internal/config/config.go
```diff
@@ -0,0 +1,145 @@
...
+import (
+	"io"
+	"os"
+
+	"github.com/BurntSushi/toml"
+	"github.com/fatih/color"
+	"gopkg.in/go-playground/validator.v9"
+)
+
+type tomlConfig struct {
```
RadhiFadlillah: I'm not really a fan of this. Shiori is intended to be a single file app with zero configuration to make it easy to install and distribute. Hence why it uses Go and SQLite by default.

However, in some case configuration file is a necessity, just like in this LDAP case. With that said, IMHO it would be better if the config file only contained the LDAP part.

internal/config/config.go
```diff
@@ -0,0 +1,145 @@
...
+	return &config
+}
+
+func init() {
+
+	cWarning := color.New(color.FgHiYellow)
+
+	config = newTomlConfig()
+
+	file, err := os.Open("shiori.toml")
```
RadhiFadlillah: I don't think it's good to parse the config file in `init`. First, IMHO error prone method like `os.Open` shouldn't be run on init. Second, since it in init, we can't give the user freedom to specify custom name or path for the config file.

ynsta:
Hello,

Thanks for your review.

Yes reusing ldap connections yes I think could improve this part, I'll look to do it soon.

For the config, I took a similar approach to the one used in traefik, if no configuration is provided runs with default settings but when a config is present takes information from it.

I also think that having different configurations in different places is confusing, so I put all configuration elements in it, and not only for the server application. Maybe init is still not the best place to put the parse, maybe it should be made lazy by loading the config the first time when accessed, so if the config is not used no need to parse it.

To finish you might need in the future to provide some other settings to the user without having him to give back arguments in each docker run/exec commands.

For testing purposes I will submit a docker compose with openldap when I'll have some time to work on it.

PR #159, opened 2019-09-22
Add code of conduct

0xsuid:
- Added Code of conduct
- Added RadhiFadlillah's email for getting reports


RadhiFadlillah:
I agree we need an CoC, however I don't really like the CoC from Contributor Covenant because IMHO it's too long. I prefer shorter CoC like Ruby's which is short and concise yet still right on target.

With that said, I'll close this issue and put modified Ruby's CoC instead.

PR #170, opened 2019-09-24
Initial PostgreSQL support

bvp:


RadhiFadlillah:
Hey, thanks !

bvp:
Thank you ;)
Need to update wiki

nfam:
Hi @bvp, thank you for PostgreSQL support. I love this DB.
However I could not view an individual bookmark (eg. Bookmark not found). Reviewing `func (db *PGDatabase) GetBookmark()` in the file `shiori/internal/database/pg.go`, I found SQL syntax error.

```sql
SELECT
	id, url, title, excerpt, author, public, 
	content, html, modified, content <> "" has_content
FROM bookmark WHERE id = $1
```

I think it should be single quote `''` instead of double route `""`, so it should be

```sql
SELECT
	id, url, title, excerpt, author, public, 
	content, html, modified, content <> '' has_content
FROM bookmark WHERE id = $1
```

RadhiFadlillah:
@nfam nice catch, thanks. It has been fixed in [9ba9d1e](https://github.com/go-shiori/shiori/commit/9ba9d1eced3417c96402bec1b3310a9e8629c6a3).

PR #205, opened 2019-10-14
Fix: grid layout on dialog

### LINKED ISSUE #204: Layout on add new account broken in Safari
The spacing between the username and password fields is too big in Safari (12.1.2):

Screenshot 2019-10-14 at 00 01 09

Pressing OK after entering the user name (expecting to enter the pw on the next screen) gives: Error: "ERROR Password must not empty"

nfam:
Fix #204 
This happens on Add Bookmark and Tags dialogs as well.

Cause: The default value of `align-content` is `stretch` while `height` of `.custom-dialog-body` is unspecified, thus, the dialog body height will be calculated from browser viewport height.

Solution: `align-content: start` will help Safari able to determine the dialog body height from its children.

Tested on Safari, Firefox on macOS. Please help test on Windows.


RadhiFadlillah:
Hey, thanks !

It works in Firefox on Linux as well.

PR #209, opened 2019-10-28
Inherit user id/group for created directories and files

rcribbs:
I use shiori on my server in docker, but I prefer to lock down the shared mount to a user on the host server.
As it stands shiori just always creates everything under root so even if it's run under another user it doesn't play well in creating files/directories.

These changes are only in the core and in the initial data directory creation. I imagine there are some other parts of the code that could use this treatment for consistency as well, but I'm only using these portions so I only addressed these portions of the code.

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -59,6 +59,12 @@ func preRunRootHandler(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
+	err = os.Chown(dataDir, os.Getuid(), os.Getgid())
+	if err != nil {
+		cError.Printf("Failed to chown data dir: %v\n", err)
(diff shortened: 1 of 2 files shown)


deanishe:
Hi @rcribbs, I've taken over the maintenance of Shiori, and I'm working through the open PRs first.

Could you explain what your PR does because I don't really understand what it's designed to achieve. Is this something specific to Docker? When I run Shiori under the user `shiori` all the files it creates also belong to that user.

rcribbs:
It's been some time since I opened this PR so I just re-tested it. It seems to be working now. Before there were some files that weren't being properly owned. It's possible a change in the code, a change in go, or a change in the docker container fixed this. I'm not sure, but I'm happy enough to just call it working.

PR #233, opened 2020-03-12
use goroutine to handle heavy process

hulb:
When I use web extension in firefox, I find that it's a little slow to get a success response when I click the save button. So I made a little change of the api to make it fast.

### DIFF
--- a/internal/webserver/handler-api-ext.go
+++ b/internal/webserver/handler-api-ext.go
@@ -34,76 +34,87 @@ func (h *handler) apiInsertViaExtension(w http.ResponseWriter, r *http.Request, 
 
 	// Check if bookmark already exists.
 	book, exist := h.DB.GetBookmark(0, request.URL)
-
-	// If it already exists, we need to set ID and tags.
-	if exist {
-		book.HTML = request.HTML
-
-		mapOldTags := map[string]model.Tag{}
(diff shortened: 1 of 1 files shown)


deanishe:
Hi @hulb, thanks for the PR.

I don't think this is the right way to fix slowness in the extension. The problem is that your change makes it impossible to inform the client if fetching the bookmark failed (and that's the part most likely to fail).

Not notifying the client that an operation it performed failed is awfully bad behaviour because it's very confusing for the user.

The problem of the extension and Shiori frontend blocking while a bookmark is being added should be fixed in the extension/frontend code, not the backend, imo.

If we start performing client-initiated operations asynchronously in other goroutines, then we need to add some way for the client to check whether they succeeded or not. And that would be a lot more complicated than updating the extension/frontend to not block the whole UI while a bookmark is being added.

hulb:
Yes, I'm almost agree with you. I also think there should be a task queue behinde to process the task asynchronously  and there would be a place to check whether the task is done. I don't think the extension/frontend  shoule keep waiting until the process done especially in the worse network condition.

deanishe:
Waiting isn't the same as blocking. The frontend/extension can wait for a response without blocking the whole UI until it's finished.

hulb:
Yes, that's exactly what I want.

PR #238, opened 2020-03-29
[+] release for several platforms

juev:
Hello,

I would like to provide configuration github actions for build shiori for several platforms.

It's used matrix for build and publish.

For this github action you shoud provide your github token in your secrets with name TOKEN.
If you cannot provide this variable build was broken.

Github Action is running whtn you create new release.

How this release it look like:
https://github.com/juev/shiori/releases/tag/v1.5.3


juev:
I'm checked macos release after build, it worked.

deanishe:
Hi @jeuv, thanks for the PR. I've taken over maintenance of Shiori and I'm working through existing PRs.

Could you rebase on master/resolve the conflict?

This PR looks very useful. I have a few questions/requests:

1. What kind of token do I need to add to the repo secrets? A personal access token?
2. Is there a particular reason the action installs p7zip via Homebrew? Does `/usr/bin/zip` not work?
3. Is there a particular advantage to using `-ldflags="-s -w"`? I know it makes the binaries a bit smaller, but it also strips debugging information, which is useful, isn't it?
4. To be on the safe side, you should probably run `go generate ./...` before building to make sure that the bundled assets are up to date.

fmartingr:
Hello from the future @juev, thanks a lot for this contribution! There has been some movement the past few weeks on this and it's actually implemented by #365! I'm closing the PR (and sad to do so for a first timer) but if you feel like improving the current state of the CI a new PR will be welcome as well. Thank you!

PR #253, opened 2020-05-29
Adopt the new APIs from go-app-paths

panxl:


fmartingr:
This has been updated on #345 with a general dependency update. Thank you for your contribution!

PR #255, opened 2020-06-20
fix the function use error

wang700:
Signed-off-by: wang700 <wanger.xyz@gmail.com>


wang700:
https://github.com/go-shiori/shiori/pull/253 already fixes this. Close this one.

PR #264, opened 2020-08-11
bug: Docker build

lapwat:
Dockerfile is now using local source instead of downloading the sources from github.

Build is now working.


symgryph:
I ran this it builds fine. Tests worked.

deanishe:
Thanks very much.

PR #285, opened 2020-12-05
Add support for search bar to search tags without "tag:"

adrianknorr:
It was not possible to search for a tag without the prefix "tag:".
For a better usage, it would be nice to search tags without the
prefix.
With the prefix, it only searches the tags, like before.

### DIFF
--- a/internal/database/sqlite.go
+++ b/internal/database/sqlite.go
@@ -240,11 +240,19 @@ func (db *SQLiteDatabase) GetBookmarks(opts GetBookmarksOptions) ([]model.Bookma
 		query += ` AND (b.url LIKE ? OR b.excerpt LIKE ? OR b.id IN (
 			SELECT docid id 
 			FROM bookmark_content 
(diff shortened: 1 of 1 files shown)


fmartingr:
Hey @adrianknorr , thanks for the PR!

I'm not really sure about this though. Since the search already works with the `tag:` prefix and your PR only modifies the behaviour for one database engine (users of postgresql/mariadb will still use the `tag:` prefix) I'm currently against this kind of change. Also we lose the ability to actually filter by tags only this way (since it will look for the word in the text and in the tags).

Also, we could implement some kind of "negation" for filters (like, `tag:readlater notag:read` to fetch items with one tag but not another).

Again, I'm not 100% sure about his and I would love to hear from the community the usage cases of the search. Maybe we could open a Feature Request regarding the search usage to gather ideas and decide on which features do users want?

**Edit:** Nevermind, I already opened it: https://github.com/go-shiori/shiori/discussions/375

Hope you understand, and I hope this won't push you away from making other contributions!

PR #290, opened 2021-01-14
improve first time usage, password-less

cauerego:
this will needs to be better documented, but at least it ain't freeze users from testing and using the software. plus it'll work great for my use case, "achoz". 😘

### DIFF
--- a/internal/view/login.html
+++ b/internal/view/login.html
@@ -78,12 +78,6 @@
                     // https://github.com/facebook/react/issues/1159#issuecomment-506584346
                     this.username = document.querySelector('#username').value;
                     this.password = document.querySelector('#password').value;
-					// Validate input
-					if (this.username === "") {
-						this.error = "Username must not empty";
-						return;
-					}
-
 					// Remove old cookie
 					document.cookie = `session-id=; Path=${new URL(document.baseURI).pathname}; Expires=Thu, 01 Jan 1970 00:00:00 GMT;`;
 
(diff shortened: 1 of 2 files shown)


cauerego:
added a commit there that improves it a bit further. now it should be just pressing login button, nothing else needed, for first use case.

fmartingr:
Hey @cauerego, thanks for the contribution! Sadly I believe this is out of scope, for various reasons:

- I plan to remove the default "account" (which isn't an account) since all user related data belongs in the database.
- In the future, a default user could be automatically populated by a migration
- Also in the future, the CLI could have account management to easily create the first user account
- Having this that way is removing a required check in the frontend.
 - In order to have a proper "single-user" shiori, we need to properly discuss it in #352

Honestly I'm sad closing PRs for contributors (and even more for first timers) but we need to think what to implement here first in order for it to be easy and useful for everyone.

cauerego:
@fmartingr sounds like a perfect decision to me.

the whole point of this was to actually remove users from shiori to begin with. i guess it just wasn't sure clear at first. 🤣

yes, let's discuss it there on #352. 😏

IcedQuinn:
Henno. It seems 1.5 (the version in Void repos at time of posting) still has the account system. And you have to dig in to the documentation to find what the default name/password is.

The CLI should really just tell you what this is when you initialize the database.

fmartingr:
> Henno. It seems 1.5 (the version in Void repos at time of posting) still has the account system. And you have to dig in to the documentation to find what the default name/password is.
> 
> The CLI should really just tell you what this is when you initialize the database.

Honestly this seems like a really good first iteration: let users know in the logs what the default account credentials are and that they should change them. Opening an issue about it.

PR #296, opened 2021-01-25
Add PWA service worker, manifest, and assets

m-bers:
Not a perfect lighthouse score, but Shiori is installable as a PWA, and the app currently works offline for articles that have already loaded. 
Screen Shot 2021-01-24 at 8 40 27 PM

Either use localhost or serve (with SSL/TLS) using a reverse proxy like NGINX:

```
server {
    listen       80;
    listen  [::]:80;
    server_name  shiori.example.com;
    return 301 https://shiori.example.com$request_uri;
}

server {
    listen              443 ssl;
    server_name         shiori.mbe.rs;

    ssl_protocols       TLSv1 TLSv1.1 TLSv1.2;
    ssl_ciphers         HIGH:!aNULL:!MD5;

    ssl_certificate     /etc/letsencrypt/live/example.com/cert.pem;
    ssl_certificate_key /etc/letsencrypt/live/example.com/privkey.pem;

    location / {
        proxy_pass http://shiori:8080;
    }
}
```

The goal is to eventually make use of the [Web Share Target API](https://web.dev/web-share-target/) so it's as easy to save a link to Shiori from Android/iOS as it is with the browser extension.


fmartingr:
Hey @m-bers, thanks a lot for this contribution! 

Not sure if you noticed but @geekgonecrazy have continued your work on #325, so I'm going to maintain that PR and mention you there so you can take a more expert look if you don't mind.

PR #297, opened 2021-02-03
Add GET endpoint to extension API

hursey013:
Allows checking if a bookmark already exists from the extension by using:

`https://shiori.domain/api/bookmarks/ext?value={{encoded URL}}`

This paves the way to allow the extension to see if a bookmark exists without relying on a local browser bookmark folder.

### DIFF
--- a/internal/view/login.html
+++ b/internal/view/login.html
@@ -1,148 +1,202 @@
 <!DOCTYPE html>
 <html lang="en">
+  <head>
+    <base href="$$.$$" />
+    <title>Login - Shiori</title>
 
-<head>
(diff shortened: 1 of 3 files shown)


fmartingr:
Hey @hursey013 , thanks for the PR! 

A few notes before moving forward:

- Rebase the changes from latest master, since the session duration was already implemented by #446 the changes in the assets and login files are no longer required.
- We can use status codes instead of payloads. So, instead of returning a `bool` we can just use the `200` and `204` status codes.

Please let me know if you're (or not) interested in moving this forward so I can prioritise it myself.

PR #306, opened 2021-05-07
Use environment variables to set the default user

soulteary:
This PR made some minor adjustments to make it easier for docker users to use.

Allow users to use environment variables to set the account used to run the program for the first time.

You can even use the initial account directly, so it can be used out of the box.

### DIFF
--- a/internal/webserver/handler-api.go
+++ b/internal/webserver/handler-api.go
@@ -74,9 +74,19 @@ func (h *handler) apiLogin(w http.ResponseWriter, r *http.Request, ps httprouter
 	accounts, err := h.DB.GetAccounts(searchOptions)
 	checkError(err)
 
-	if len(accounts) == 0 && request.Username == "shiori" && request.Password == "gopher" {
+	defaultUser, _ := os.LookupEnv("SHIORI_DEFAULT_USER")
+	defaultPass, _ := os.LookupEnv("SHIORI_DEFAULT_PASS")
+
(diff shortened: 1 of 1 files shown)


braderhart:
Might I suggest setting a password file environment variable as well:

https://docs.docker.com/engine/swarm/secrets/

soulteary:
> https://docs.docker.com/engine/swarm/secrets/

Thank you for your reply.

If you use secrets, you will rely on swarm mode. 

Obviously, simply using env will be more common.

Maybe one more option should be added if there are many swarm users. :)

soulteary:
> @soulteary Secrets aren't exclusive to Swarm and can be used with just Docker compose.

Thanks for the reminder, I will check it out later.

fmartingr:
Hey @soulteary, thanks for this PR. Sadly I'm doing a cleanup and I'm going to consider this out of scope at the moment, and plan to maintain it that way until proper account management is introduced in the CLI and the _shiori pseudo-user_ can be removed from the application code.

Quoting myself from https://github.com/go-shiori/shiori/pull/290#issuecomment-1042072827:

> - I plan to remove the default "account" (which isn't an account) since all user related data belongs in the database.
> - In the future, a default user could be automatically populated by a migration
> - Also in the future, the CLI could have account management to easily create the first user account
>  - In order to have a proper "single-user" shiori, we need to properly discuss it in #352

PR #322, opened 2021-11-18
Add a basic Docker compose file

Sleuth56:
I noticed you didn't have a Docker compose file. So I figured sense I wrote one for my install I would contribute it back so others can use it. It's not complex at all. Just gets it up and running.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,10 @@
+version: "3"
+services:
+  shiori:
+    image: radhifadlillah/shiori:latest                  
+    container_name: Shiori
+    ports:
+      - "8080:8080"
+    volumes: 
+      - ./container:/srv/shiori
+    restart: unless-stopped


fmartingr:
Hello @Sleuth56, thanks for your contribution! 

I'm sorry to say that having a compose file is not necessary right now since you can just as easily fire up shiori using a single line command, while the docker-compose has to make assumptions as to where to put volumes, publish ports and so on. Also real-life scenarios will use additional services and networks that I do not want to cover in the main repository (since we support different database backends). I think is better for users to tune up their compose files to each one needs.

Just as a reminder, if you still use Shiori, [that we migrated the container images to a new registry](https://github.com/go-shiori/shiori/discussions/373) here in Github.

amitjethani:
This is quite helpful especially for less technical folks relying on Portainer or deploying to their NAS. Thank you @Sleuth56

fmartingr:
Hey @amitjethani, I actually run my containers through portainer and have no issue using that whatsoever. The problem of having a compose file directly here is that we have to maintain it and there are some questions related on how people run their's services. Should we maintain three different compose files for each database backend? How could we parametrise variables, secrets, volume names and/or paths, etc? I personally don't use that for development so I would end up not updating it if anything changes (and would need to, and keep an eye on PR's for the same).

Since this is a request, I would be happy to make a `shiori-docker-compose` repository available for the community to hop on and maintain. How does that sound to you?

amitjethani:
My perspective is that just a basic compose file is all that's needed as a "quick start" (defaulting to sqlite). Power users who need Postgres or MySQL, or need to set up a Traefik Proxy, can probably author their own.

Re: parameterizing variables: placeholder names is all that's needed imo. It just gives folks a quick primer on how to get going and all they have to do is manually substitute out a path volume or something: `${REPLACE_ME}/shiori:/srv/shiori`

Leave it up to the folks at linuxserver.io to pick up the project and provide examples for more complex use-cases (they might have a submission process?)

PR #341, opened 2022-02-06
Remove golang from final docker image

n8225:
The current Dockerfile uses the golang:alpine image for the final image, which makes the final image much larger than needed.  This changes the final image to alpine and uses images from ghcr.io(It looks like these are ghcr.io official images pulled from dockerhub).  Also added a label to the image that will link the package to the repository.


fmartingr:
Thank you @n8225 ! I've already merged and pushed the new image to the GHCR.

Issue #254, opened 2020-06-07
Panic while trying to run cross compiled binary on Raspberry PI


quaintdev:
Cross compiled binary for Raspberry PI panics on device with following error 

```
panic: Binary was compiled with 'CGO_ENABLED=0', go-sqlite3 requires cgo to work.

goroutine 1 [running]:
github.com/jmoiron/sqlx.MustConnect(...)
	/home/user/go/pkg/mod/github.com/jmoiron/sqlx@v1.2.0/sqlx.go:650
github.com/go-shiori/shiori/internal/database.OpenSQLiteDatabase(0x2776000, 0x26, 0x0, 0x0, 0x0)
	/home/user/go/src/github.com/go-shiori/shiori/internal/database/sqlite.go:23 +0x384
github.com/go-shiori/shiori/internal/cmd.openSQLiteDatabase(0x5834d9, 0xb, 0x0, 0x0)
	/home/user/go/src/github.com/go-shiori/shiori/internal/cmd/root.go:116 +0x70
github.com/go-shiori/shiori/internal/cmd.openDatabase(0x27600a0, 0x1c, 0x1ff, 0x0)
	/home/user/go/src/github.com/go-shiori/shiori/internal/cmd/root.go:110 +0x74
github.com/go-shiori/shiori/internal/cmd.preRunRootHandler(0x2768b40, 0xb716b8, 0x0, 0x0)
	/home/user/go/src/github.com/go-shiori/shiori/internal/cmd/root.go:63 +0x174
github.com/spf13/cobra.(*Command).execute(0x2768b40, 0xb716b8, 0x0, 0x0, 0x2768b40, 0xb716b8)
	/home/user/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:810 +0x17c
github.com/spf13/cobra.(*Command).ExecuteC(0x2768000, 0x0, 0x0, 0x46324)
	/home/user/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:914 +0x230
github.com/spf13/cobra.(*Command).Execute(...)
	/home/user/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:864
main.main()
	/home/user/go/src/github.com/go-shiori/shiori/main.go:19 +0x18
```

deanishe:
You can't cross-compile Shiori, I'm afraid. cgo doesn't support cross-compilation, and SQLite requires cgo.

[...]

PR #345, opened 2022-02-07
Update Dependencies and Replace mattn/go-sqlite with modernc.org/sqlite

n8225:
This PR replaces mattn/go-sqlite with modernc.org/sqlite, updates all dependencies, and brings go version to 1.17
Static binaries can now be built without cgo.
In my limited testing everything appears to be working.  It probable needs some more thorough testing.

#254 #253


n8225:
I will hopefully be done redoing the whole CI/CD process in the next few days.  If you want I can remove the older go version checks in this PR.

fmartingr:
Also, can you verify that the shiori data dir does not change with the library upgrades? I did this on a commit in my fork and the data dir changed from `~/.local/share/shiori/shiori.db` to `~/.local/share/shiori/shiori/shiori.db`. Thanks!

n8225:
> Also, can you verify that the shiori data dir does not change with the library upgrades? I did this on a commit in my fork and the data dir changed from `~/.local/share/shiori/shiori.db` to `~/.local/share/shiori/shiori/shiori.db`. Thanks!

I verified the data dir is still `~/.local/share/shiori/shiori.db`

PR #363, opened 2022-02-10
fix: use shiori binary as container entrypoint

### LINKED ISSUE #318: modify the default port 8080 on Docker
Is it possible to modify the default port 8080 to other on the docker image? 

I tried: 

```
docker run --rm radhifadlillah/shiori serve -h
docker run --rm radhifadlillah/shiori serve --port 8081
```

But I get an error: 

```
docker: Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "serve": executable file not found in $PATH: unknown.
```

fmartingr:
Using shiori as entrypoint would allow image users to add arguments to the executable call, using their own flags or calling one-shot containers to perform maintenance/manipulation tasks.

What do you think @n8225 ?

Closes #318

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -20,3 +20,9 @@ markComment: >
   Thank you for your contributions.
 # Comment to post when closing a stale issue. Set to `false` to disable
 closeComment: false
+
+pulls:
+  daysUntilStale: 10
+  daysUntilClose: 10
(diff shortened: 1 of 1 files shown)


n8225:
Makes sense to me.  Makes it easier to use the shiori cli with docker.

Issue #364, opened 2022-02-11
Each time a bookmark is updated it adds a duplicate content entry


fmartingr:
I had a problem with my database and had to start from a backup. I added a few of my missing bookmarks but forgot to add the tags, so when I edited them,  it started showing up as duplicates in the frontend.

<img width="1995" alt="Screenshot 2022-02-11 at 19 52 00" src="https://user-images.githubusercontent.com/812088/153652284-17ad3e1a-c6d5-448c-8a4b-cc120da2fdb2.png">

This is a backend issue, since I checked the XHR requests and it showed up several times in the list, gladly with the same ID meaning its not duplicated in the database so it may be a problem with the SQL?

<img width="727" alt="Screenshot 2022-02-11 at 19 52 59" src="https://user-images.githubusercontent.com/812088/153652403-45f26bec-6798-43b9-b879-9d3837b61b13.png">

This **also happen when searching**.

fmartingr:
Offending code: https://github.com/go-shiori/shiori/blob/2ca628b7fe8f6bc86cb2367a2ed964a1e9f083db/internal/database/sqlite.go#L160-L161

fmartingr:
Will be fixed in next release, manual SQL to remove duplicates from `bookmarks_content`: 

``` sql
DELETE FROM bookmark_content
WHERE rowid NOT IN (
    SELECT min(rowid)
    FROM bookmark_content
    GROUP BY docid
)   
```

PR #367, opened 2022-02-13
fix: update bookmark on sqlite database

fmartingr:
It used an update and insert at the same time, since the virtual table
cannot have constrants nor upserts we need to try an update the
bookmark content to check for it's existence, inserting it only if the
update fails (don't affect any rows).

This will need to be improved in the future, since not all bookmark
updates require modification of the content table (in my case, I was
testing only updating tags).

Closes: #364


fmartingr:
```

PR #388, opened 2022-02-28
Add Go version for heroku

My6UoT9:
Heroku requires an extra line so it knows which go version to use.
https://github.com/heroku/heroku-buildpack-go#go-module-specifics

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -1,0 +1,6 @@
+[metadata.heroku]
+  root-package = "github.com/My6UoT9/shiori"
+  go-version = "go1.14.1"
+  install = [ "." ]
+  ensure = "reuw"
+  additional-tools = ["github.com/golang-migrate/migrate"]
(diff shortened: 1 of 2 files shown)


fmartingr:
Hey @My6UoT9, thanks for the contribution! I'm merging this right away. 

I may put a "Deploy to heroku" button in the readme as well, that still exists, right?

fmartingr:
> I think it does yes, I am sorry I should have tested more extensively. Sofar it seems I loose all data when heroku starts again, I have to check how to have data persisted.

Yeah, I assume that using SQLite the database will be created on the dyno directly, which is ephemeral by default. If I recall correctly, in order to add heroku support properly an `app.json` needs to be created in the root of the repository. So far your change may be useful for people that want to take a quick peek at Shiori, they can deploy it to heroku and test it there even though a more permanent solution will require more work.

My6UoT9:
Hm yes, maybe add a warning :)

PR #397, opened 2022-03-07
[FEATURE] Sanitize strings before writing to the database

imajes:
This pulls the sanitizing elements from PR #250, to make this (and that PR) easier to move forward. Thanks to @ynsta for the original work. :D

### DIFF
--- a/internal/database/database.go
+++ b/internal/database/database.go
@@ -2,6 +2,8 @@ package database
 
 import (
 	"database/sql"
+	"strings"
+	"unicode/utf8"
 
 	"github.com/go-shiori/shiori/internal/model"
 )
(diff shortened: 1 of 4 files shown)


fmartingr:
Hey @imajes ,thanks for your work on this! 

Please give me a bit to check this and #250 locally so I can better understand what the problem is and the best course of action. Since this involves the database and there are some work to be done in this area I'd rather not merge everything at once.

fmartingr:
Hey @imajes, can you please provide with a test content to check what runes give an error? So far I've tried with latin, japanese, korean and emoji and the sanitize funcion does not alter them in any way. Maybe the postgresql database for Shiori isn't created with UTF-8 encoding by default?

fmartingr:
Please read my comments at #250

Issue #385, opened 2022-02-25
Adding a URL doesn't automatically create an archive


thelazyoxymoron:
I have turned on the "Create archive by default" setting and I see the relevant option ticked as well when I'm adding a new URL. However, it doesn't actually create an archive, I have to manually go and do update archive.

Grauwolf:
when using SQLite it seems to work as intended. when configuring a PostgreSQL database, I can confirm that no archive is created, even though "create archive" is enabled.

fmartingr:
> I have turned on the "Create archive by default" setting and I see the relevant option ticked as well when I'm adding a new URL. However, it doesn't actually create an archive, I have to manually go and do update archive.

Is this happening to you in postgresql as well?

MidnightSnowleopard:
I can confirm I am seeing the same behavior with PostgreSQL.  Haven't test for SQLite but the behavior is present on Shiori 1.5.2 and 1.5.2 in docker.

thelazyoxymoron:
> Is this happening to you in postgresql as well?

Sorry, should've mentioned this, I'm using Postgresql, haven't tried with SQLite.

fukajun:
Maybe I fixed it. Please review PR #480.

fmartingr:
Apparently fixed in #484, please open a new issue if you experience something similar again.

PR #418, opened 2022-04-06
fix: Adding a URL doesn't automatically create an archive (#385)

ogumaru:
Fixed a bug pages not archived although createArchive checkbox checked. (fix: #385)

"Create archive" checkbox was not applied.

image

### DIFF
--- a/internal/webserver/handler-api.go
+++ b/internal/webserver/handler-api.go
@@ -307,7 +307,7 @@ func (h *handler) apiInsertBookmark(w http.ResponseWriter, r *http.Request, ps h
 		book.Title = book.URL
 	}
 
-	if !payload.Async {
+	if payload.CreateArchive {
 		book, err = downloadBookmarkContent(book, h.DataDir, r)
 		if err != nil {
 			log.Printf("error downloading boorkmark: %s", err)


fmartingr:
Hey @ogumaru, I've checked this and is working as intended. 

When you add a link it'll show up empty (only the link) at first, but the content and archive will be fetched in the background, just wait a few seconds (depending on the content length) and if you refresh you will see that the bookmark will have content/archive if extraction was possible.

If you are experiencing issues with this under PostgreSQL (#418) I didn't have the time to check that just yet (I'm pretty busy these weeks, but thanks a lot for checking it out) but this PR breaks the async behaviour of the endpoint (and also fetches the content twice if Async is `true`, which is the default).

This requires more research on why is only happening with one database engine and not the other (I'm running the latest development container image over SQLite on my end with no issues, so at a quick glance this may be database and not API related).

ogumaru:
@fmartingr 
Thank you checking despite you are so busy, and sorry for my lack of understanding. 
> archive will be fetched in the background

I misunderstood this behavior, I had assumed downloading not started.
(I prefer this async behavior, thank you nice updating.)

> fetches the content twice if Async is true

This is also my misunderstood. I'm not familiar with golang, so I had assumed `go func()` is a function declaration.

Sorry for not helping, thank you.

fmartingr:
Hey @ogumaru, I didn't mean to be rude! I'm always grateful for people to get their hands dirty by trying to fix things, I'm just one person after all! Anyway, I'm glad to know that Shiori is working fine on your end.

PR #422, opened 2022-05-03
implement migration for mysql & postgres

hulb:


fmartingr:
Hey @hulb; now that the SQLite migrations were working as expected this was the obvious next step, so I'm very grateful you took the time to implement this yourself.

hulb:
@fmartingr you're welcom. I'm glad to make some help. I have used shiori for a long time, it helps a lot.

PR #425, opened 2022-05-10
Improve performance for sqlite queries

### LINKED ISSUE #408: Performance regression
I recently updated Shiori from the old 1.5.0 release to the latest version and noticed a substantial performance regression. I'm using sqlite for the database (with around 4500 bookmarks) and the performance was perfectly fine with version 1.5.0.

When running the latest code (which includes commit [sqlite: remove unneeded LEFT JOIN to improve performance](https://github.com/go-shiori/shiori/commit/ca3cc11b2de4d894e5fee3e1deea86ccfb541093) ), just loading the bookmarks list after logging in takes about 20 seconds (and causes very high cpu load on the server during those 20 seconds). Switching to the next page takes similarily long. In comparison, on version 1.5.0 the same thing took less than a second.

I noticed that migrating the database to FTS5 caused its size to almost double (395 MB vs. 202 MB). I don't know if this is related in any way, but I thought it might be worth mentioning.

Orhideous:
This PR is intended to address a range of performance regressions for the SQLite database.

Namely:
- Removed the slow JOIN for bookmark lists
- Reduced the number of queries for tags

Tested manually on 3k bookmarks with content. Last version ([7394b10](https://github.com/go-shiori/shiori/commit/7394b10)) returned main page in a second or two. Now it takes few, maybe tens of *milliseconds*.

Note that migration is required: `shiori migrate`.

Built an image for the test, now I use it myself: `docker.io/orhideous/go-shiori:1.5.2-perf-fix`

_P.S. There is definitely a room for improvement for another databases._

Fixes #408

### DIFF
--- a/internal/database/migrations/sqlite/0002_denormalize_content.down.sql
+++ b/internal/database/migrations/sqlite/0002_denormalize_content.down.sql
@@ -1,0 +1,16 @@
+BEGIN TRANSACTION;
+CREATE TABLE IF NOT EXISTS bookmark_temp(
+    id INTEGER NOT NULL,
+    url TEXT NOT NULL,
+    title TEXT NOT NULL,
+    excerpt TEXT NOT NULL DEFAULT '',
+    author TEXT NOT NULL DEFAULT '',
+    public INTEGER NOT NULL DEFAULT 0,
+    modified TEXT NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    CONSTRAINT bookmark_PK PRIMARY KEY(id),
+    CONSTRAINT bookmark_url_UNIQUE UNIQUE(url)
+);
+INSERT INTO bookmark_temp SELECT id, url, title, excerpt, author, public, modified FROM bookmark;
+DROP TABLE bookmark;
+ALTER TABLE bookmark_temp RENAME TO bookmark;
+COMMIT;
--- a/internal/database/migrations/sqlite/0002_denormalize_content.down.sql
+++ b/internal/database/migrations/sqlite/0002_denormalize_content.down.sql
@@ -1,16 +1,3 @@
 BEGIN TRANSACTION;
-CREATE TABLE IF NOT EXISTS bookmark_temp(
-    id INTEGER NOT NULL,
-    url TEXT NOT NULL,
-    title TEXT NOT NULL,
-    excerpt TEXT NOT NULL DEFAULT '',
-    author TEXT NOT NULL DEFAULT '',
-    public INTEGER NOT NULL DEFAULT 0,
-    modified TEXT NOT NULL DEFAULT CURRENT_TIMESTAMP,
-    CONSTRAINT bookmark_PK PRIMARY KEY(id),
-    CONSTRAINT bookmark_url_UNIQUE UNIQUE(url)
-);
-INSERT INTO bookmark_temp SELECT id, url, title, excerpt, author, public, modified FROM bookmark;
-DROP TABLE bookmark;
-ALTER TABLE bookmark_temp RENAME TO bookmark;
+ALTER TABLE bookmark DROP COLUMN has_content;
 COMMIT;
--- a/internal/database/migrations/sqlite/0002_denormalize_content.up.sql
+++ b/internal/database/migrations/sqlite/0002_denormalize_content.up.sql
@@ -1,0 +1,6 @@
+ALTER TABLE bookmark
+    ADD has_content BOOLEAN DEFAULT FALSE NOT NULL;
+
+UPDATE bookmark
+SET has_content = bc.has_content FROM (SELECT docid, content <> '' AS has_content FROM bookmark_content) AS bc
+WHERE bookmark.id = bc.docid;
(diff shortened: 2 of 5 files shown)


internal/database/migrations/sqlite/0002_denormalize_content.up.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE bookmark
```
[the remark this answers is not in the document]
fmartingr: We should also wrap this with `BEGIN ... TRANSACTION`, I missed that in the first migration as well. 😓

internal/database/sqlite.go
```diff
@@ -302,25 +307,75 @@ func (db *SQLiteDatabase) GetBookmarks(opts GetBookmarksOptions) ([]model.Bookma
...
+	// If content needed, fetch it separately
+	// It's faster than join with virtual table
+	if opts.WithContent {
+		contents := make([]bookmarkContent, 0, len(bookmarks))
+		contentMap := make(map[int]bookmarkContent, len(bookmarks))
+
+		contentQuery, args, err := sqlx.In(`SELECT docid, content, html FROM bookmark_content WHERE docid IN (?)`, bookmarkIds)
+		contentQuery = db.Rebind(contentQuery)
+		if err != nil {
+			return nil, fmt.Errorf("failed to expand query: %v", err)
```
fmartingr, line 325:
```suggestion
			return nil, fmt.Errorf("failed to expand bookmark_content query: %v", err)
```

internal/database/sqlite.go
```diff
@@ -302,25 +307,75 @@ func (db *SQLiteDatabase) GetBookmarks(opts GetBookmarksOptions) ([]model.Bookma
...
 		FROM bookmark_tag bt
 		LEFT JOIN tag t ON bt.tag_id = t.id
-		WHERE bt.bookmark_id = ?
-		ORDER BY t.name`)
+		WHERE bt.bookmark_id IN (?)
+		ORDER BY t.name`, bookmarkIds)
+	tagsQuery = db.Rebind(tagsQuery)
 	if err != nil {
-		return nil, fmt.Errorf("failed to prepare tag query: %v", err)
+		return nil, fmt.Errorf("failed to expand query: %v", err)
```
fmartingr, line 358:
```suggestion
		return nil, fmt.Errorf("failed to expand bookmark_tag query: %v", err)
```

fmartingr (commented):
Hey @Orhideous, **amazing work!** This is just as I imagined it, and it should speed up a lot the bookmark browsing for people with huge databases, thanks a lot!

I've left a few minor comments in the PR, please take a look once you have time.

After we have this running we may need to apply the same denormalization to the `hasArchive` to stop trying to look for files in the filesystem only to check if something exists. Or even maybe migrating the content to the database as well, I don't know yet.

Again, thanks a lot for the contribution!

Orhideous:
@fmartingr Thanks for the review and feedback. I'll commit suggestions ASAP.

BTW, while testing this fix on real data, I found several problems indirectly related to this performance problem, and some security issues. This PR should be more polished, IMHO. Nothing serious, but it needs to be refined.

fmartingr:
No problem @Orhideous, just let me know when this is ready to review again. If you found more stuff to check, either open it in a new PR if it's not related to this changes or reference your findings in a issue so we can keep an eye on them.

Orhideous:
@fmartingr I think it's quite ready.

internal/database/migrations/sqlite/0002_denormalize_content.down.sql
```diff
@@ -0,0 +1,16 @@
+BEGIN TRANSACTION;
+CREATE TABLE IF NOT EXISTS bookmark_temp(
```
fmartingr, line 2: Why didn't you just `ALTER TABLE bookmark DROP COLUMN has_content;` ?

fmartingr (commented):
Hey @Orhideous, this looks amazing! I've made a comment in the down migration since SQLite should support the `DROP COLUMN` statement.

After that, this is ready to go into the dev channel for people to test out there!

Awesome work, thanks a lot!

internal/database/migrations/sqlite/0002_denormalize_content.down.sql
Orhideous: I used workaround from old times when DROP COLUMN wasn't supported in sqlite. Hold a sec, I'll fix it.

Orhideous: @fmartingr It still not supported, alas. So I leave this migration as-is.

fmartingr: @Orhideous Support [was added in 3.35.0](https://www.sqlite.org/releaselog/3_35_0.html) and our library should support [up to 3.38.3](https://gitlab.com/cznic/sqlite/-/commit/61483fe9bc293b715fd476c7396b45f345544954). What error are you experiencing?

Orhideous: Oh, that explains a lot. Focal Fossa distro still uses 3.31.1. Okay, thanks for clarification!

fmartingr (approved):

internal/database/migrations/sqlite/0002_denormalize_content.down.sql
fmartingr: Awesome, let's go!

PR #436, opened 2022-06-17
add disable-auth for serve command to avoid login, useful when serve shiori behind SSO (like authelia)

chain710:
- use disable-auth to avoid login, useful when serve shiori behind SSO (like authelia)


fmartingr:
Hey @chain710, thanks for the PR!

I believe a better implementation for this would be leaving the login endpoint as it is, but if the flag is provided (which I would rename to `single-user` to follow #352) it would:
- Check if the database has at least one admin user:
  - If not, create the gopher/shiori user in the database
  - If it only has one, login that user directly (to allow for people that already has their user set up)
  - If it has more than one, return an error and prevent usage

What do you think?

chain710:
Thank you for the advice. This takes more work, I'll look to it later

Issue #445, opened 2022-06-28
Server error with CSS files with new version


pgrunm:
Hello, I'm running Shiori with my Raspberry pi behind a Caddy reverse proxy. I followed the guide [here](https://github.com/go-shiori/shiori/pull/374#issuecomment-1044704202) to run Shiori behind Caddy as reverse proxy. 

Until the last update everything was fine, but since 1.5.3 any CSS file isn't loaded correctly if you run Shiori like this shiori serve --webroot /shiori/ parameter. In Firefox I can see internal server errors (HTTP code 500) and this is also displayed on the console but no further information. 

![image](https://user-images.githubusercontent.com/8317771/176247663-d58fe684-c83e-469a-89af-b57dd849e640.png)

The page as result looks like this now:

![image](https://user-images.githubusercontent.com/8317771/176108818-f3b1e7b2-b8fb-4ee4-b777-950eba255deb.png)

And now the great part: If you run Shiori without the --webroot /shiori/ parameter, everything is fine. I also verified this on my local machine, once with the parameter and once without, same result as with the Raspberry Pi, without reverse proxy of course.

## Steps to reproduce

- Download Shiori (if not already in path or something)
- Extract and run with to test: shiori serve --webroot /shiori/ and open up Shiori http://127.0.0.1:8080/shiori/. This is how no CSS is displayed.
- To verify the working state run Shiori like this: shiori serve and open Shiori again.
- If I try to open the `source-sans-pro.min.css` file as an example my browser says the following: `open /css/source-sans-pro.min.css: invalid name`. The corresponding log entry from the containers log is: 
`time="2022-06-28T17:43:40Z" level=warning msg="GET /shiori/css/source-sans-pro.min.css" proto=HTTP/1.1 remote="172.19.0.2:43460" reqlen=0 size=48 status=500`

hulb:
Sorry for your trouble, it will be fixed when the PR merged

pgrunm:
No problem, I appreciate your upcoming fix. :-)

fmartingr:
Hey @pgrunm, I just merged @hulb's fix. Please check if this solves it, if not, please reopen the issue so we can take a look.

pgrunm:
Works great, thank you very much! 👍

PR #448, opened 2022-07-06
fix invalid assets path when use webroot

hulb:
fix https://github.com/go-shiori/shiori/issues/445

last time I used embed.Fs to replace the assets generation, it turns out that the sub path is difference. For exmaple, When using assets generation, the `css` assets would be at `/css`, and when using embed.Fs it must be `css` since the embed.Fs doesn't allow sub path start with `/`: https://github.com/golang/go/blob/master/src/io/fs/fs.go#L41

### DIFF
--- a/internal/webserver/handler-ui.go
+++ b/internal/webserver/handler-ui.go
@@ -14,9 +14,10 @@ import (
 	"strings"
 
 	"github.com/PuerkitoBio/goquery"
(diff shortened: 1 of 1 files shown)


fmartingr:
Hey @hulb, thanks a lot for your contributions to Shiori and your work on this fix!

PR #451, opened 2022-07-10
Fix docker migration script example

kaimoe:
The example docker command given was invalid, as `docker run` operates on images, not containers. This is the closest working alternative using the same command, with another option being `docker exec containername shiori migrate` to operate on an existing container.


fmartingr:
You're totally right here @kaimoe, thanks for fixing the reference and providing a much more complete command, appreciated!

PR #452, opened 2022-07-25
fixbug: the bookmark Primary key also auto increment when insert false.

Schr0dingerCat:
fixbug: the bookmark Primary key also auto increment when insert false.so after upsert or insert, we need get the right bookmark id with the url. so you are not sure the bookmark id before you insert sucessfully into bookmark table. you shoud get the id after insert sucessfully


fmartingr:
Hey @Schr0dingerCat, thanks for this! Sadly this will be fixed when #484 get merged, which fixes a related bug and uses the proper ID when inserting. If I missunderstod the problem you're trying to solve please reach out again so we can properly issue this and reopen the PR.

Issue #466, opened 2022-08-10
Next API iteration


fmartingr:
### Discussed in https://github.com/go-shiori/shiori/discussions/462

<div type='discussions-op-text'>

<sup>Originally posted by **Orhideous** August  8, 2022</sup>
Hi everyone.

A few months ago I started to sketch out some complete API overhaul (see: https://github.com/go-shiori/shiori/pull/463). The main objectives that I set for myself were:
- simplify the code base for easier support and long-term maintainability;
- leverage battle-tested technologies and approaches as not to reinvent the wheel;
- add ability to generate any client based solely on [OpenAPI specification](https://github.com/Orhideous/shiori/blob/feature/api-overhaul/openapi.yml) and thus get rid of the routine of writing it manually.

All that is left for me now is just to copy the business logic to the [handlers](https://github.com/Orhideous/shiori/tree/feature/api-overhaul/internal/api).

So far I already integrated [oapi-codegen](https://github.com/deepmap/oapi-codegen/) into codebase. I intentionally left out the old APIs untouched so that they would work in parallel for some time (at least one release). Also, I intentionally designed new API as close as possible to the old one.
This will give time for a more or less painless migration.

@fmartingr What do you think about it? I would be happy to get feedback.</div>

PR #467, opened 2022-08-10
draft: logic domains

fmartingr:
A draft pull request to explain and experiment with logic domains, to try and encapsulate the common code between the API and the CLI prior to API improvements (#466)


fmartingr:
Continuing work in #497

PR #472, opened 2022-08-21
docs: fix remember represented as a number instead of a bool

babariviere:
Changed in pull request: https://github.com/go-shiori/shiori/pull/346

Since the API does not support numbers for the `remember` field in `/api/login`, I changed the documentation to make it a boolean instead of a number.


fmartingr:
Hey @babariviere, good catch! Thank you for your contribution.

PR #480, opened 2022-09-28
Fix bug when add new bookmark using PostgreSQL

fukajun:
Fixed a bug pages not archived although createArchive checkbox checked. (fix: https://github.com/go-shiori/shiori/issues/385) 

Maybe it is only occurred using postgres.
To be precise, the archive is saved, but the ID to refer to different.

This ploblem reason is diffrent create file path id(archive and thumb) and create bookmark record id.

Function of CreatNewID return recently numbered bookmark id that is not same to be insert bookmark record id.
Returned bookmark id is used to path of save thumb and archive.
But, record inserted by next numbered id.
As a result, New bookmark record try to refered  not exists directory pathes. 

This fix will be changes CreateNewID returned to be next numbered bookmark record id.

Excuse, This implements is not think about concurrency bookmarking.
But, Current implement is same about this point.

### DIFF
--- a/internal/database/pg.go
+++ b/internal/database/pg.go
@@ -607,7 +607,7 @@ func (db *PGDatabase) RenameTag(id int, newName string) error {
 // CreateNewID creates new ID for specified table
 func (db *PGDatabase) CreateNewID(table string) (int, error) {
 	var tableID int
-	query := fmt.Sprintf(`SELECT last_value from %s_id_seq;`, table)
+	query := fmt.Sprintf(`SELECT last_value + 1 from %s_id_seq;`, table)
 
 	err := db.Get(&tableID, query)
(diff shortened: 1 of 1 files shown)


fmartingr:
Hey @fukajun, thank you for finding this bug! While the solution is not optimal, the actual implementation is not very good as well, so I'm good with merging this into Shiori, but could you add some tests so we have some workflows covered? I have plans to refactor this in the future so we don't need to get a new ID from the database (we will either generate ir ourselves or insert the bookmark before creating archives).

fmartingr:
Hey @fukajun I have taken the liberty of solving this on #452 along with other related issue. Thank you for reporting this and taking the time to provide a fix, really appreciated!

PR #481, opened 2022-09-30
Draft: Replace WARC with Obelisk

### LINKED ISSUE #353: Support Obelisk archiving
It seems that shiori depends on [warc](https://github.com/go-shiori/warc) which is currently archived. We need to find a replacement for warc. Maybe [obelisk](https://github.com/go-shiori/obelisk)?

## Acceptance criteria
- Add a migration that will define in which archiver type the content is (put `warc` for already existing rows, but `obelisk` as default)
- Add logic to allow multiple archivers to be used, **do not** remove Warc logic, just refactor it.
- Allow the `/bookmark/:id/archive` handler to load multiple archive types (to load old and new)
- Allow the `POST /api/v1/bookmarks`/`POST /api/v1/bookmarks/cache`/`POST /api/v1/bookmarks/:id/cache` to select which archiver to use (but hardcode/default it to `obelisk`).
- Add a documentation page describing the archivers, available options, pro-cons.
- Determine if different extensions should be used from now on (leave current filename expectations **intact**)
- All code logic should be properly tested
- Swagger documentation should be updated

fmartingr:
This is a draft to replace WARC with Obelisk in order to know what would be the impact of such a change.

Notes:
- Serving content (not archives) uses the archive in some way. Don't know why, need to investigate
- Old archives would require a migration from the old system (WARC) to the new one (Obelisk) potentially saving up space and being much more confortable to manage since it will create single files.
- Right now storage is file-based, I would move things to database with a cache layer so its easier for us to manage and for users to backup, manage and migrate.

Closes #353

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/disintegration/imaging v1.6.2
 	github.com/fatih/color v1.13.0
 	github.com/go-shiori/go-readability v0.0.0-20220215145315-dd6828d2f09b
+	github.com/go-shiori/obelisk v0.0.0-20220524135250-3d6752a59bd7
 	github.com/go-shiori/warc v0.0.0-20200621032813-359908319d1d
 	github.com/go-sql-driver/mysql v1.6.0
 	github.com/gofrs/uuid v4.2.0+incompatible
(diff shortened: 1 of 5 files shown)


Katarn:
> Right now storage is file-based, I would move things to database with a cache layer so its easier for us to manage and for users to backup, manage and migrate.

I would make this functionality optional. I'm afraid to imagine how large the database dumps will be, they are clearly not intended for such use.

hulb:
Actually I have a need for shiori to save these archive files to object storage like minio :)

fmartingr:
> Actually I have a need for shiori to save these archive files to object storage like minio :)

I believe you already can mount volumes as s3-compatible storage by using docker/k8s?

fmartingr:
This was very outdated and it has very fee work on it. The PDF archiver refactored most of the logic so I need to start over when that PR is merged.

Issue #276, opened 2020-09-03
There is some problem in search with chinese


zero47:
Title: macOS 免费小工具
In keyworks:
  1. macOS √
  2. 免费 √
  3. 免费小工具 ×

wadefelix:
mysql 不支持中文全文索引导致的

fmartingr:
> mariadb 不支持中文全文索引导致的

Thank you for stepping up, is this something than can be fixed or worked around application side, or it requires the user to switch to a different database engine?

natsukium:
I’m using [PGroonga](https://pgroonga.github.io/), a PostgreSQL extension, to use FTS in Japanese.
https://github.com/go-shiori/shiori/compare/master...natsukium:shiori:groonga
It also supports Chinese.
For MariaDB, there is also an extension, [Mroonga](https://mroonga.org/).

If you use the database without these extensions, you need to modify the application code like the following.
Sorry for the Japanese blog.
https://www.clear-code.com/blog/2020/12/22.html

fmartingr:
Hey @natsukium , thanks for the insights! If you are running this locally, have you noticed any caveheats running that extension? I'm unsure how many Shiori users require this in order to bundle it directly on our migrations or just provide a separate documentation page for people to install it on their own. Bundling it directly would require supporting it in all database systems, since want to provide the same features regardless of the database backend if possible.

PR #494, opened 2022-10-09
fix err in keyword searching

### LINKED ISSUE #470: Search error: failed to fetch count: pq: syntax error at or near "AGAINST" (500)
Version: 
ghcr.io/go-shiori/shiori:v1.5.3-6-g3deeda7
postgres:14.5-alpine3.16

log:
```
time="2022-08-17T05:14:11Z" level=warning msg="GET /api/bookmarks?keyword=ndt&tags=&exclude=&page=1" proto=HTTP/1.1 remote="10.42.1.25:45580" reqlen=0 size=61 status=500
```
```
2022-08-17 05:14:11.273 UTC [7045] ERROR:  syntax error at or near "AGAINST" at character 101
2022-08-17 05:14:11.273 UTC [7045] STATEMENT:  SELECT COUNT(id) FROM bookmark WHERE TRUE AND (
				url LIKE $1 OR
				MATCH(title, excerpt, content) AGAINST ($2 IN BOOLEAN MODE)
			)
```

### LINKED ISSUE #477: Error when searching for a keyword that does not exist in the database
Just try typing in the search field something like "asdjalksdjlasjd".

> failed to expand bookmark_tag query: empty slice passed to 'in' query (500)

I use `sqlite` backend. Apparently it happens here:

https://github.com/go-shiori/shiori/blob/3deeda7151b8d719df2d9b7ac74cdcad16aa805e/internal/database/sqlite.go#L351-L359

And apparently the `bookmarkIds` array is empty.

Shiori v1.5.3.

hulb:
fix #470 #477

### DIFF
--- a/internal/database/pg.go
+++ b/internal/database/pg.go
@@ -223,7 +223,9 @@ func (db *PGDatabase) GetBookmarks(ctx context.Context, opts GetBookmarksOptions
 	if opts.Keyword != "" {
 		query += ` AND (
 			url LIKE :lkw OR
-			MATCH(title, excerpt, content) AGAINST (:kw IN BOOLEAN MODE)
+			title LIKE :kw OR
+			excerpt LIKE :kw OR
+			content LIKE :kw
 		)`
 
 		arg["lkw"] = "%" + opts.Keyword + "%"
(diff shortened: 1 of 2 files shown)


hulb:
fix #470 #477

internal/database/pg.go
```diff
@@ -223,7 +223,9 @@ func (db *PGDatabase) GetBookmarks(ctx context.Context, opts GetBookmarksOptions
 	if opts.Keyword != "" {
 		query += ` AND (
 			url LIKE :lkw OR
-			MATCH(title, excerpt, content) AGAINST (:kw IN BOOLEAN MODE)
```
hulb, line 226: It seems that MATCH ... AGAINST is a mysql syntax for text searching. In postgres, I don't find such simple way to implement text matching so I use LIKE multi times as a workround.

fmartingr: Yeah, we need to implement FTS into PSQL, I need to take a look at #276 first in order to decide an engine.

fmartingr:
Hey @hulb, as always, thank you for fixing this! Little by little we're making Shiori more stable and usable :)

PR #495, opened 2022-10-09
panic if err occurs in downloading bookmark content when insert bookmark asynchronously

### LINKED ISSUE #459: Crash when bookmark URL is not accessible
## Step to reproduce:

1. Start up `shiori serve` and login
2. Add a bookmark that is not accessible for shiori
    1. If the bookmark is likely i/o timeout, need to click the `Update archive` on the saved bookmark, the options `Keep the old title and excerpt` and `Update archive as well` is not necessary to check or uncheck.
    2. If the bookmark is likely `connection refused`, the shiori will crash immediately.
4. shiori crash

## Log is following:

```
2022/08/04 08:57:14 error downloading boorkmark: error downloading bookmark: Get "http://127.0.0.1:2323": dial tcp 127.0.0.1:2323: connect: connection refused
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xaaf64a]

goroutine 68 [running]:
github.com/go-shiori/shiori/internal/webserver.(*handler).apiInsertBookmark.func1()
        /home/runner/work/shiori/shiori/internal/webserver/handler-api.go:329 +0x8a
created by github.com/go-shiori/shiori/internal/webserver.(*handler).apiInsertBookmark
        /home/runner/work/shiori/shiori/internal/webserver/handler-api.go:324 +0x5c5
```

The results is same when the error message is `i/o timeout`, `read: connection reset by peer` or `connection refused` like above.

## Version

`shiori_1.5.3_Linux_x86_64.tar.gz` from GitHub release page.

hulb:
fix #459

### DIFF
--- a/internal/webserver/handler-api.go
+++ b/internal/webserver/handler-api.go
@@ -323,7 +323,7 @@ func (h *handler) apiInsertBookmark(w http.ResponseWriter, r *http.Request, ps h
 	if !payload.Async {
 		book, err = downloadBookmarkContent(book, h.DataDir, r)
 		if err != nil {
-			log.Printf("error downloading boorkmark: %s", err)
+			panic(fmt.Errorf("failed to save bookmark: %v", err))
 		}
 	}
 
@@ -338,6 +338,7 @@ func (h *handler) apiInsertBookmark(w http.ResponseWriter, r *http.Request, ps h
 			bookmark, err := downloadBookmarkContent(book, h.DataDir, r)
 			if err != nil {
 				log.Printf("error downloading boorkmark: %s", err)
+				return
 			}
 			if _, err := h.DB.SaveBookmarks(ctx, *bookmark); err != nil {
 				log.Printf("failed to save bookmark: %s", err)


internal/webserver/handler-api.go
```diff
@@ -323,7 +323,7 @@ func (h *handler) apiInsertBookmark(w http.ResponseWriter, r *http.Request, ps h
 	if !payload.Async {
 		book, err = downloadBookmarkContent(book, h.DataDir, r)
 		if err != nil {
-			log.Printf("error downloading boorkmark: %s", err)
+			panic(fmt.Errorf("failed to save bookmark: %v", err))
```
fmartingr, line 326: I don't like this approach, mostly because:
1) I don't like the usage of panic/recover, it makes following errors complicated for contributors.
2) We don't want to stop here if the download fails, we just need to report the error and save the bookmark anyway, since the user can trigger a manual download on its own.

hulb: Yes, I agree. #513 is a better solution.

PR #496, opened 2022-10-09
replace http route with go-chi

hulb:
I prefer go-chi as a light route because I think it's more idiomatic and makes the code more understandable

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,13 +7,13 @@ require (
 	github.com/PuerkitoBio/goquery v1.8.0
 	github.com/disintegration/imaging v1.6.2
 	github.com/fatih/color v1.13.0
+	github.com/go-chi/chi/v5 v5.0.7
 	github.com/go-shiori/go-readability v0.0.0-20220215145315-dd6828d2f09b
 	github.com/go-shiori/warc v0.0.0-20200621032813-359908319d1d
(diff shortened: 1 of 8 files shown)


hulb:
I'm also hesitating about whether using [go-gin](https://github.com/gin-gonic/gin) is a better choice.

fmartingr:
I'm going to leave this on hold, since a refactor of the API has to be made but I'd rather go in parallel to the current one. I was about to go with [fiber](https://gofiber.io/) since [I already used it and have some experience and QoL with it](https://github.dev/fmartingr/notion2ical/tree/latest/internal/server/http/server.go), but the engine doesn't matter much to me. Any comments towards gin or other http engines?

hulb:
No,using an engine will make me to have to understand the rules of the engine.I can try fiber later.

fmartingr:
I had the same opinion until I saw how much boilerplate I had to avoid coding 😅 You just need to figure out how to properly organize the code, the rest should be easy. Please take a look at #497. I will open a discussion so we can comment on this.

fmartingr:
Discussion here: https://github.com/go-shiori/shiori/discussions/498

hulb:
I have reviewed #497 , it's a big step to next.I think this PR is unneccesary and just close it.

PR #497, opened 2022-10-09
draft: next API iteration

fmartingr:
Added a proof of concept (extending the idea from #467) with a new API server built on [fiber](https://gofiber.io).

- Added a separate `server` command to launch the new HTTP server
- Database interface remains the same, though it should change a little bit, but it works for the PoC.
- Added logic Domain which will encapsulate logic related to shiori domains, to avoid having duplicated code between the client, the api and the extension API (which will merge)

Open to comments!


internal/http/routes/api/auth.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+type AuthAPIRoutes struct {
+	logger *zap.Logger
+	router *fiber.App
+	deps   *config.Dependencies
+}
+
+func (r *AuthAPIRoutes) Setup() *AuthAPIRoutes {
+	r.router.Post("/login", r.loginHandler)
```
hulb, line 20: I think it's a little veiled to declare route here. I prefer the original way, put routes together and use difference handlers

internal/domains/auth.go
```diff
@@ -0,0 +1,56 @@
...
+	"time"
+
+	"github.com/go-shiori/shiori/internal/database"
+	"github.com/go-shiori/shiori/internal/model"
+	"github.com/golang-jwt/jwt/v4"
+	"go.uber.org/zap"
+	"golang.org/x/crypto/bcrypt"
+)
+
+type AuthDomain struct {
```
hulb, line 15: It's a good idea to use domain to seperate common login between cli and web api. Is the name `domain` a usual one? I usually use `service`.

internal/config/dependencies.go
```diff
@@ -0,0 +1,22 @@
+package config
+
+import (
+	"github.com/go-shiori/shiori/internal/database"
+	"github.com/go-shiori/shiori/internal/domains"
+	"go.uber.org/zap"
+)
+
+type Dependencies struct {
```
hulb, line 9: I don't think it's proper to put this `Dependencies` here under the `config` dir. And I think it's a little early to put dependencies in a struct. How about we do these things later when enough dependencies occur?

internal/config/config.go
```diff
@@ -0,0 +1,91 @@
+package config
+
+import (
+	"bufio"
+	"context"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/sethvargo/go-envconfig"
```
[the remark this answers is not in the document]
fmartingr: Right now we have a mix of environment variables and CLI flags. I tend to use environment variables because I only use the server version, but we need to figure out what's the best solution for both worlds. Ideally the CLI would read configuration from some file as well. I will take a look at Viper, it seems to provide everything on one package :)

internal/http/response/response.go
```diff
@@ -0,0 +1,27 @@
+package response
+
+import "github.com/gofiber/fiber/v2"
+
+type Response struct {
+	Ok          bool              `json:"ok"`
+	Message     interface{}       `json:"message"`
+	ErrorParams map[string]string `json:"error_params,omitempty"`
+	statusCode  int
```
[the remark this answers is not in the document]
fmartingr: This status code is the one used for the HTTP response. I should clarify that in the comments.

internal/domains/auth.go
fmartingr: I tend to use `domain` when `service` can be missinterpreted or used for other things, so I just went with it. Right now I think is irrelevant and I could change it before going forward with this.

internal/http/routes/api/auth.go
fmartingr: The problem with using a single package for everything is that after a while you either have huge files containing all kinds of related logic for the API (handlers, payloads, responses, middlewares, codecs, ...). I followed a similar approach here as with the "domains", each endpoint is contained in a "domain", in this case this is the "auth" domain, so all routes for the auth domain are defined here, and all domains are defined in _api.go_. (auth, bookmarks, tags, ...). As with everything, suggestions welcome!

internal/http/server.go
```diff
@@ -0,0 +1,99 @@
...
+		})).
+		Use(middleware.NewZapMiddleware(middleware.ZapMiddlewareConfig{
+			Logger:      s.logger,
+			CacheHeader: "X-Cache",
+		})).
+		Use(func(c *fiber.Ctx) error {
+			return c.Next()
+		}).
+		Use(recover.New()).
+		Mount(cfg.Routes.System.Path, routes.NewSystemRoutes(s.logger, cfg).Setup().Router()).
```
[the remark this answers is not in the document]
fmartingr: So, I I understood you correctly. Would you define all routes here directly?

internal/domains/auth.go
hulb: I agree. The name `service` is used everywhere now.

internal/http/routes/api/auth.go
hulb: I get you.I think it's just a matter of taste. I don't think I'm much experienced in dev in go. I can learn from this :)

internal/http/server.go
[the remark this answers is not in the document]
hulb: Yes. If you decide to declare every route in each domain, that is fine. As for here, I just think it's not elegant engouh. Too much words to code And I'm a little lazy :)

internal/http/response/response.go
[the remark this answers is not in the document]
hulb: Oh, I misunderstood the purpose of `Response`.It's good.

internal/config/dependencies.go
hulb: Yeah, when there is many dependencies, it's better to do like this. I just think maybe it can be added in the future when really needed.Actually I mean if the signature like this:
```
func NewAuthAPIRoutes(logger *zap.Logger, db database.DB, authDomain domains.AuthDomain) *AuthAPIRoutes {
```
it seems more straightforward since there are just two dependencies for AuthAPI.

fmartingr: Yeah, maybe I tried looking too far into the future. I will change it to your approach using the simpler dependency injection pattern.

PR #499, opened 2022-10-11
fix `failed to save bookmark: context canceled` and `failed to save bookmark: ID must not be empty`

Acelya-9028:
[fix: failed to save bookmark: context canceled](https://github.com/go-shiori/shiori/commit/f12bfd014932741a58b2b0eba9400a32aeb944e5)

When the HTTP request completes, the asynchronous bookmarks save may be incomplete. In this case, the shared context is canceled and saving bookmarks fails.

Introduced by https://github.com/go-shiori/shiori/commit/09f246506587afe60dfa61bda680ea0c4fa9d36b

[fix: failed to save bookmark: ID must not be empty](https://github.com/go-shiori/shiori/commit/4c0092fb73966584cb5cc9740cb2e111c55fc878)

SQLite and MySQL fail to save a bookmark because the database does not return the last inserted id.

This commit fix only SQLite so MySQL is still broken.

Introduced by https://github.com/go-shiori/shiori/commit/222fc14fe29f9988db9aca2972c2f3a29c3914b0 and https://github.com/go-shiori/shiori/commit/dc73cd825b9d8b997364062b625bb4b3f15bc1c8


Acelya-9028:
Done :)

fmartingr:
Thank you for your contribution!

PR #512, opened 2022-10-12
feat: add migrate flag to allow users to migrate in all calls to the CLI

### LINKED ISSUE #473: First start creates an empty database with no tables
I run:

```bash
./shiori --portable serve
```

And I see a zero size file `shiori.db` in the `shiori-data` directory:

```bash
ls -l shiori-data
-rw-r--r-- 1 user users 0 aug 23 21:57 shiori.db
```

Open in browser, try to login `shiori`/`gopher` and see:

```
failed to fetch accounts: SQL logic error: no such table: account (1) (500)
```

It works only if before the first start run:

```bash
./shiori --portable migrate
```

I use shiori v1.5.3.

fmartingr:
Added the `--migrate` flag (alongside the `migrate` command) to perform migrations automatically on each call to the CLI that has that flag defined.

I'm reluctant to perform it in background directly, so this is a good compromise on both sides.

Fixes #473

### DIFF
--- a/docs/Usage.md
+++ b/docs/Usage.md
@@ -3,9 +3,9 @@ Before using `shiori`, make sure it has been installed on your system. By defaul
 <!-- TOC -->
 
 - [Running Docker Container](#running-docker-container)
-- [Using Command Line Interface](#using-command-line-interface)
-    - [Search syntax](#search-syntax)
 - [Running migrations](#running-migrations)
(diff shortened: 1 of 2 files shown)


fmartingr:
This issue has been closed due to feedback received in #473. After discussion with the community we're going back to automatic migrations through the application, which would require appropriate testing on the database engines, which is already in the works and we need it anyway.

PR #514, opened 2022-10-12
fix(db): avoid using dash in fts5 queries

### LINKED ISSUE #492: Using dash in search field doesn't work and gives error
- Database: SQlite
- Search input: `what-happens-when`

Screen Shot 2022-10-08 at 1 04 42 PM

fmartingr:
Replaces the dash (`-`) in keyword for the FTS5 arguments in SQLite, since that's a column token for the engine, so queries coming from the search with a dash where trying to find a column where the user would just want a keyword match.

Added some tests to handle this behavior.

Fixes #492

### DIFF
--- a/internal/database/database_test.go
+++ b/internal/database/database_test.go
@@ -2,6 +2,7 @@ package database
 
 import (
 	"context"
+	"log"
 	"testing"
 
 	"github.com/go-shiori/shiori/internal/model"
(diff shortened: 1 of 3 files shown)


internal/database/database_test.go
```diff
@@ -16,8 +16,10 @@ func testDatabase(t *testing.T, dbFactory testDatabaseFactory) {
 		"testCreateBookmark":              testCreateBookmark,
 		"testCreateBookmarkTwice":         testCreateBookmarkTwice,
 		"testCreateBookmarkWithTag":       testCreateBookmarkWithTag,
-		"testCreateTwoDifferentBookmarks": testCreateTwoDifferentBookmarks,
 		"testUpdateBookmark":              testUpdateBookmark,
+		"testGetBookmarks":                testGetBookmarks,
+		"testGetBookmarksCount":           testGetBookmarksCount,
+    "testCreateTwoDifferentBookmarks": testCreateTwoDifferentBookmarks,
```
Katarn, line 22: Here, apparently, it was necessary to use tabs.

PR #515, opened 2022-10-13
fix: failed to save bookmark: constraint failed: UNIQUE constraint failed: bookmark.url

Acelya-9028:
After a first bookmark save, all subsequent ones fail because the update query does not include an identifier and the update is done on all rows.

Add a test to create two different bookmarks.

Introduced by 05fee53bd0b7e81b5e74c4ae9efcbe213b563744


fmartingr:
This was a big oversight on my part, thanks a lot for te catch @Acelya-9028 ! Merging right away.

Acelya-9028:
Thank you very much :)

PR #525, opened 2022-10-19
feat: add option to bypass document readability check

Acelya-9028:
A check is done to test if a content can be cached but this test may disqualify a readable document so an option is needed to allow users to bypass this test and force bookmarks processing.

Try to fix https://github.com/go-shiori/go-readability/issues/26


fmartingr (changes requested):
Hey @Acelya-9028, thanks for the submission! I will take a look at https://github.com/go-shiori/go-readability/issues/26 as well, though I think I haven't experienced this.

This looks good to me, the only thing I would want is some test to check that this works properly, we are lacking a lot in that area so is better if we start adding tests while we work on the changes. Can you take care of that?

internal/view/js/component/dialog.js
```diff
@@ -109,27 +109,31 @@ export default {
 		fields: {
 			immediate: true,
 			handler() {
-				this.formFields = this.fields.map(field => {
-					if (typeof field === 'string') return {
+				this.formFields = this.fields.reduce((fields, field) => {
```
fmartingr, line 112: Why this change? Is this related to the PR?

Acelya-9028:
Hey @fmartingr Ok for tests, I will add this

internal/view/js/component/dialog.js
Acelya-9028: Yes, it's because `if (field.visible === false) return` in a `map` insert a null in the array and Vue don't like this. If my memory is good, it's the reason why I replaced the `map` by a `reduce`, to filter null.

So why there seemed to not have problem with `if (typeof field === 'string') return` before? Maybe `typeof field` is always equal to string but I have not verified I admit.

Acelya-9028: Because I add a new checkbox to force the caching only if the caching previously failed. If the caching worked great, no need to force it so no need to display it I think.

However, it's all the time displayed is multiple bookmarks are selected.

PR #527, opened 2022-10-24
Remove typo in compose Dockerfile post-run command

AIndoria:
This should be `serve` not `server` 

server was causing it to freak out about the typo.


fmartingr (approved):
Thank you @AIndoria ! This is my fault, since I created this compose for #497 😓 
Thank you for fixing this.

PR #550, opened 2023-01-05
Fix #549 reset page when searching for new tag

### LINKED ISSUE #549: Clicking on tag shows no bookmarks in search
## Data
- **Shiori version**: 1.5.3
- **Database Engine**: SQLite
- **CLI/Web interface/Web Extension**: Web interface

When clicking on a tag on a bookmark that is in a page > 1, it results in the search returning no results.

## Expected behavior
Bookmarks are returned in results.

## To Reproduce
1. Add more than 30 bookmarks (or however many bookmarks required to create more than 1 page).
2. Add a tag to the last bookmark on page 2.
3. Click the tag on the bookmark on page 2
4. It will add the 'tag:my-tag' filter to search and show no results

## Notes
This issue is caused because the page query parameter is not stripped when clicking on a tag on a bookmark.

If you are on page 2 the url is as follows:

`/?page=2`

Click on tag e.g. 'webdev'.

`/?page=2&search=tag%3Awebdev`

However there is only 1 bookmark and hence page 2 is empty.

Probably just needs to set `this.page = 1;` here:
https://github.com/go-shiori/shiori/blob/a950bfec4c75eadbe5c1a03302ba9875a8236d45/internal/view/js/page/home.js#L295-L297

I can create a PR if it is an acceptable fix.

Thanks,
Ryan

ryanwr:
Fix issue #549 

Tested fix and working as expected now.

FIxes #549

### DIFF
--- a/internal/view/js/page/home.js
+++ b/internal/view/js/page/home.js
@@ -305,6 +305,7 @@ export default {
 
 			if (tagName === "*") {
 				this.search = excludeMode ? "-tag:*" : "tag:*";
+				this.page = 1;
 				this.loadData();
 				return;
 			}
@@ -344,6 +345,7 @@ export default {
 				}
 			}
 
+			this.page = 1;
 			this.loadData();
 		},
 		showDialogAdd() {


fmartingr:
Hey @ryanwr, thanks a lot for taking the time and fixing that! Merging it right away. Will cut a release later in the week, meanwhile we can test drive this in the development image.

PR #554, opened 2023-01-12
feat: improve bookmark image detection

Katarn:
Unfortunately, not all web servers are configured correctly. I found a [detailed comment on Stack Overflow](https://stackoverflow.com/a/54488403) from which it becomes clear that the MIME type `image/jpg` does not exist. But I am faced with the fact that on some servers, however, this type is specified. And as a result, the cover does not load.

Here is an example of a link: https://vc.ru/581066

```bash
$ curl -I -X GET "https://vc.ru/cover/fb/c/581066/1673501221/cover.jpg"

HTTP/2 200
content-type: image/jpg
```

Before and after this change:

изображение

### DIFF
--- a/internal/core/processing.go
+++ b/internal/core/processing.go
@@ -159,7 +159,11 @@ func downloadBookImage(url, dstPath string) error {
 
 	// Make sure it's JPG or PNG image
 	cp := resp.Header.Get("Content-Type")
-	if !strings.Contains(cp, "image/jpeg") && !strings.Contains(cp, "image/png") {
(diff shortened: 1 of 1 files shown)


fmartingr:
Hey @Katarn, thank you for your investigation and patch to address this issue, it'll make the bookmark list so much prettier. The only downside is that processing all bookmarks again is a no-go for people like me with hundreds already in the database.

Katarn:
@fmartingr I think users will still need to update their bookmarks after moving to the obelisk. :)

PR #559, opened 2023-01-25
fix migration error for mysql when STRICT_TRANS_TABLES enabled

renothing:
https://dev.mysql.com/doc/refman/8.0/en/data-type-defaults.html#data-type-defaults-explicit


fmartingr:
Hey @renothing, thanks for your contribution. I'm assuming this comes from the fact that `TEXT` types cannot have a default value [^1]? if you could add a little description to your PR so we can have it explained. Thank you!

[^1]: https://dev.mysql.com/doc/refman/8.0/en/blob.html#:~:text=BLOB%20and%20TEXT%20columns%20cannot%20have%20DEFAULT%20values

PR #560, opened 2023-01-28
fix: browser ext cannot archive and update bookmark.

IvanLi-CN:
After I changed the code, I noticed that next-api (#497) seems to have fixed the problem, so you might consider merging this pr if next-api is not released in time.


fmartingr:
Hey @IvanLi-CN, thanks for taking care of this. I'm going to merge it since it will take some time to get the new API up and running.

PR #561, opened 2023-01-29
fix(frontend): avoid night mode flickering on reload

sdvcrx:
Hi!

Shiori pages are flickering on refresh(or initialize), you can watch the video below.

https://user-images.githubusercontent.com/5170232/215329259-f1d02b9a-3899-480e-bb44-3038a23f3a13.mp4

This patch can fix it by moving nightMode related code to `<head>`


fmartingr:
Hey @sdvcrx, thanks for the fix. I'm just wondering if there could be a more elegant way of handling this? Does Vue have any other event we could use to avoid having this logic in two places at once?

PR #562, opened 2023-01-29
fix(frontend): prevent archive header covering up the archive content

sdvcrx:
image


fmartingr:
hey @sdvcrx, thanks for taking the time to fix this. One little request though, could you make use of less variables so we have more control over it?

Adding `@header-height: 60px;` to the top of the file, using that on the header height itself and then on the margin we could just use `@header-height + 1px` should suffice if my less knowledge is not very rusty.

Thanks again!

sdvcrx:
@fmartingr Please take a look again :)

fmartingr:
> @fmartingr Please take a look again :)

Thank you for the changes, that's exactly what I was referring to :)

PR #565, opened 2023-01-31
Add log args for root command

chain710:
For improved debugging convenience, add some log flags in root command

- log-level: specify log level, default info
- log-caller: whether log `file:lineno` or not, default false
- log FullTimestamp

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -2,20 +2,23 @@ package cmd
 
 import (
 	"fmt"
-	"os"
-	fp "path/filepath"
-
 	"github.com/go-shiori/shiori/internal/database"
 	"github.com/go-shiori/shiori/internal/model"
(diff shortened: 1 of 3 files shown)


internal/cmd/loglevel_value.go
```diff
@@ -0,0 +1,26 @@
+package cmd
+
+import "github.com/sirupsen/logrus"
+
+type LogLevel logrus.Level
```
fmartingr, line 5: Not saying it's wrong, but what's the benefit of having this custom type instead of parsing and validating with `logrus.ParseLevel` directly on the root command?

fmartingr (commented):
Hey @chain710, sorry for the delay! This slipped through my notifications. Thank you for the improvements on the logging, much appreciated :) I have just one question before moving forward.

fmartingr:
Thank you for the contribution @chain710 !

PR #566, opened 2023-02-01
Use UTC for cookie expiry date in backend

MoX46:
**Issue:**
Logging in to the web UI for the first time resulted in 301 redirect on v1.5.4. Only seems to be an issue on chromium based browsers (Brave, Chrome, Edge). Not an issue on Firefox as far as I can tell. 

**Solution:**
[JavaScript documents indicate cookie expires attribute should be UTC format ](https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie). After cba5046 expiry date is returned in local time format instead of UTC. Updating backed to explicitly use UTC resolves the issue.


fmartingr:
Hey @MoX46, thanks for finding this out and fixing it, really appreciated :)

PR #594, opened 2023-03-20
User auth via proxy request header

chain710:
The Auth Proxy enables user authentication using an HTTP header from an external authentication service. I propose adding two arguments in serve command:

1. reverse-proxy-auth-user: This argument specifies the name of the proxy authentication HTTP header, such as X-Auth-User.
2. trusted-proxies: This argument ensures that the `auth-proxy` is only used when requests come from trusted IP addresses.

Use case:

1. Authentication is provided by either Authelia or Keycloak.
2. Nginx passes the X-Auth-User header to Shiori, which comes from the authentication provider.
3. Shiori looks up the user in the database and returns a session.

Let me know if you have any further questions or if you'd like me to make any additional revisions.

Reference:

https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/auth-proxy/
https://github.com/gogs/gogs/issues/165
https://github.com/miniflux/v2/pull/570


cycomanic:
I'm very interested in seeing this merged. However it would be great if the arguments could also be specified via environment variables, making the feature easier to use in docker containers

fmartingr:
Hey @chain710, thanks for the contribution! Before I jump into the code reviews again, do you have any easy guidelines to check this feature locally? It would help when I clone and review this, so I can also check how it works.

I would also follow @cycomanic's comment, to make it more configurable for the admins. What do you think?

PterX:
This is a very useful feature, any progress?

fmartingr:
> This is a very useful feature, any progress?

I guess is stalled, I wanted to test it properly to prepare for a potential e2e test but didn't get any response since then. If you want to continue the work feel free to do so. Keep in mind that some things have changed since then since we are currently migrating the server to a different library, though implementing this should be easier now (I hope!)

PterX:
> > This is a very useful feature, any progress?
> 
> I guess is stalled, I wanted to test it properly to prepare for a potential e2e test but didn't get any response since then. If you want to continue the work feel free to do so. Keep in mind that some things have changed since then since we are currently migrating the server to a different library, though implementing this should be easier now (I hope!)

https://github.com/go-shiori/shiori/pull/836 I was push a pr, please check it

PR #603, opened 2023-04-18
fix(log): record user real ip from headers

Tomilla:
Fix the problem that the system log cannot record the user's real connection IP when the system is deployed in the container or behind the reverse proxy. Details of implement as follow: 
1. First, determine whether the remote addr of request is a private address. If it is a public network address, return it directly;
2. If it is private, get the real IP from `X-REAL-IP` and `X-Forwarded-For` headers in turn. ( The code handles multiple comma-separated IP lists, that is, through multi-layer reverse proxy);
3. If the above headers do not exist or are invalid, the remote addr is returned as is. 

Thank you for reviewing my code.


fmartingr:
Hey @Tomilla , thanks for the contribution! I see that you are still adding some stuff to the PR. Is this ready for review ?

Tomilla:
> Hey @Tomilla , thanks for the contribution! I see that you are still adding some stuff to the PR. Is this ready for review ?

Yes, The code can now be reviewed. 🙌 
The submission a few days ago was mainly to add unit tests to cover as many code branches as possible, and some rare cases, such as after a multiple reverse proxies, asserting some fatal panics that shall be thrown at start (i.e. defined the wrong IP CIDR). 
Although I have tried my best to make the code logic clear, if you have any suggestions for improvement, please point it out directly, and I will fix it in time (of course, during off-duty hours).

Tomilla:
It's ready for review, and I just wrote simple documentation for each function in the "utils-ip" module. :)

internal/webserver/utils-ip.go
```diff
@@ -0,0 +1,219 @@
...
+// IsPrivateIP check IPv4 or IPv6 address according to the length of byte array
+func IsPrivateIP(ip net.IP) bool {
+	if ip4 := ip.To4(); ip4 != nil {
+		return isPrivateV4(ip4)
+	}
+	return len(ip) == IPv6Len && isPrivateV6(ip)
+}
+
+// IsIpValidAndPublic is a helper function check if an IP address is valid and public.
+func IsIpValidAndPublic(ipAddr string) bool {
```
fmartingr, line 171: nit: `IsIPValidAndPublic`

internal/webserver/utils-ip.go
```diff
@@ -0,0 +1,219 @@
...
+		// We disable validations to IPs under the 6to4 anycase prefix because
+		// there's too much risk of a malicious actor advertising the prefix and
+		// answering validations for a 6to4 host they do not control.
+		// https://community.letsencrypt.org/t/problems-validating-ipv6-against-host-running-6to4/18312/9
+		parseCidr("2002::/16", "RFC 7526: 6to4 anycast prefix deprecated"),
+	}
+)
+
+// parseCidr parses the predefined CIDR to `net.IPNet` that consisting of IP and IPMask.
+func parseCidr(network string, comment string) net.IPNet {
```
fmartingr, line 134: nit: `parseCIDR`

internal/webserver/utils-ip.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	return false
+}
+
+// IsPrivateIP check IPv4 or IPv6 address according to the length of byte array
+func IsPrivateIP(ip net.IP) bool {
+	if ip4 := ip.To4(); ip4 != nil {
+		return isPrivateV4(ip4)
+	}
+	return len(ip) == IPv6Len && isPrivateV6(ip)
```
fmartingr, line 167: I think here we could just assume that is IPv6 or use `ip.To6() != nil`.

fmartingr (changes requested):
Awesome work @Tomilla. You could have easily created your own library out of this! 

I've left a few minor comments. Also if you could rename the `utils-ip` file to `utils_ip` to match with the rest of the file structure. Once that's out of the way we're ready to go :)

Thanks for the contribution!

internal/webserver/utils-ip.go
Tomilla: Ok, I will fix all code-style issues in minutes.

internal/webserver/utils-ip.go
Tomilla: It should be ip.To16() != nil, I has been fix it.

Tomilla:
Hello @ fmarting, all comments or suggestions in the above conversation have been fixed, but I do not have permission to run GitHub operations.Please try again later and I will update based on feedback.
Regarding the idea you mentioned of packaging this part of the code into a third-party library independently, I mainly consider that the functionality is still relatively limited, so I have not decided to separate it at this moment. 😄

fmartingr (approved):
Thanks for the contribution and the quick response @Tomilla ! Waiting for the CI to merge this right away.

PR #615, opened 2023-06-11
fix direction in RTL language

Monirzadeh:
Show RTL language like Persian, Arabic, etc. in correct direction.
LRT language like English, France, etc. show correctly too


fmartingr:
> I don't have any idea why checks failed. @fmartingr can you help? As I know, it is not related to my PR

I was just fixing that on `master`. Just updated your branch to see if CI works as it should.

fmartingr (approved):
Thanks for your contribution @Monirzadeh ! Really appreciated :)

PR #623, opened 2023-06-22
Initial Ebook support

Monirzadeh:
Hi @fmartingr 
as we talk in  #605 it is a not complete code. Please keep it as draft until meet all needs.
I write UI/API part too, but because I don't know you prefer of how to handle ebook status.

## What it's done:
- [x] generate an EPUB file from a readable version of bookmark.
- [x] EPUB generation done without any external library. (i prefer less dependency as possible)
- [x] it can serve EPUB file if file exist. (/bookmark/id/ebook)
- [x] not download image twice if image url duplicate in bookmark
- [x] add a generate ebook api
- [x] add style for nice view.
- [x] `hasEbook` define, and page update based on `hasEbook`.
- [x] download ebook with bookmark title
- [x] move to update archive.
- [x] use `io` instead of `io/outil` for move to higher version of go
- [x] add missing error handling while try to write inside a file
- [x] update archive regenerate ebook if it exists
- [x] Error 404 in Api instead of panic
- [x] if delet a bookmark it will remove ebook too
- [x] add link to bookmark content page. icon not add to bookmark gride like archive etc. becuse we have book icon for fast download and don't like make first page unneeded element.
- [x] remove `language` and `id` until find good solution for them. If we can find a good solution for to summarize bookmark content, we can use that for `description` tag too.
- [x] if it is PDF not generate ebook.
- [x] better CSS for ebook. i pick up css from [here](https://github.com/poire-z/crengine/blob/2b7a96c2778291d2b136abbff60715a5aa0e0925/cr3gui/data/epub.css) with very small change. I think it should be tested in long term to determine embedded CSS is good enough or not. Thanks @poire-z. @poire-z if you don't like we use this CSS here https://github.com/go-shiori/shiori/pull/623/commits/519e10d6cebe24206ce72d6f843c37a8d4ac1915, please comment me.
- [x] remove panic in all API that this PR.  ~~Just one still exist that I can't replace~~
- [x] change function name to `GenerateEbook`
- [x] [not isFatalErr anymore](https://github.com/go-shiori/shiori/pull/623/commits/eac045ef2378492ea4f967cca04343bb9071a293) - do we need to remove `isFatalErr` in other part of code like `ProcessBookmark` too?
- [x] add unit test for `GenerateEbook` 68% coverage. [this parts](https://github.com/go-shiori/shiori/pull/623#discussion_r1257323458) don't have unit test yet 
## Maybe in future not plan on this PR 
- add cover to ebook
- add language tag based on book content (maybe use an external library for that) we can use [lingua-go](https://github.com/pemistahl/lingua-go), but somehow it overkills. Or use [go-readability](https://github.com/go-shiori/go-readability) tag.
- delete action. not yet until find good solution in ui.
- return PDF in ebook download action to eink device. (is useful if eink device can handel pdf). when start working on eink device
- summarize bookmark content, and use that for `description` tag.
- Include font in ebook as font publisher. Why not do that here? It can be complicated for now because of different font for each language.
- ebook generate by default in settings. I need more time to understand how settings work in shiori.

Discussion here: https://github.com/go-shiori/shiori/discussions/605


fmartingr:
> Hi @fmartingr I don't have any idea for delete (or update) ebook in UI. is not better move this to update Archive?

I say that the update should be along the update archive/cache. We can skip delete until we have a proper UX for it. Of course it should delete if the user removes the bookmark altogether.

> Update: most of the part is done. If you think some part of code is not good enough, please report. or any specific feature

Sure, I have some time today so I'm going to review it. Thanks a lot for your hard work on this!

> Update as I do more research [go-readability](https://github.com/go-shiori/go-readability) return language of article, but shiori don't save that on database, maybe I will save that on database for add language tag. Have language of article can be helpful for better page direction text on specific situation too, not just ebook tag.

This would be helpful, yes. But same as other things, keep this PR as it is and we can iterate on it once it gets merged, so you don't add to many things at once. What do you think?

internal/core/ebook.go
```diff
@@ -0,0 +1,238 @@
...
+
+	ebookDir := fp.Join(req.DataDir, "ebook")
+	// check if directory not exsist create that
+	if _, err := os.Stat(ebookDir); os.IsNotExist(err) {
+		os.MkdirAll(ebookDir, model.DataDirPerm)
+	}
+	// create epub file
+	epubFile, err := os.Create(ebookPath)
+	if err != nil {
+		return book, true, fmt.Errorf("can't create ebook")
```
fmartingr, line 60:
```suggestion
		return book, true, errors.Wrap(err, "can't create ebook")
```

internal/core/ebook.go
```diff
@@ -0,0 +1,238 @@
...
+	defer epubFile.Close()
+
+	// Create zip archive
+	epubWriter := zip.NewWriter(epubFile)
+	defer epubWriter.Close()
+
+	// Create the mimetype file
+	mimetypeWriter, err := epubWriter.Create("mimetype")
+	if err != nil {
+		return book, true, fmt.Errorf("can't create mimetype")
```
fmartingr, line 71:
```suggestion
		return book, true, errors.Wrap(err, "can't create mimetype")
```

internal/core/ebook.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	mimetypeWriter.Write([]byte("application/epub+zip"))
+
+	// Create the container.xml file
+	containerWriter, err := epubWriter.Create("META-INF/container.xml")
+	if err != nil {
+		return book, true, fmt.Errorf("can't create container.xml")
```
fmartingr, line 78:
```suggestion
		return book, true, errors.Wrap("can't create container.xml")
```

internal/core/ebook.go
```diff
@@ -0,0 +1,238 @@
...
+	containerWriter.Write([]byte(`<?xml version="1.0" encoding="UTF-8"?>
+<container version="1.0" xmlns="urn:oasis:names:tc:opendocument:xmlns:container">
+  <rootfiles>
+	<rootfile full-path="OEBPS/content.opf" media-type="application/oebps-package+xml"/>
+  </rootfiles>
+</container>`))
+
+	contentOpfWriter, err := epubWriter.Create("OEBPS/content.opf")
+	if err != nil {
+		return book, true, fmt.Errorf("can't create content.opf")
```
fmartingr, line 90:
```suggestion
		return book, true, errors.Wrap(err, "can't create content.opf")
```

internal/core/ebook.go
```diff
@@ -0,0 +1,238 @@
...
+  </manifest>
+  <spine toc="ncx">
+    <itemref idref="content"/>
+  </spine>
+</package>`))
+
+	// Create the style.css file
+	styleWriter, err := epubWriter.Create("style.css")
+	if err != nil {
+		return book, true, fmt.Errorf("can't create content.xml")
```
fmartingr, line 112:
```suggestion
		return book, true, errors.Wrap(err, "can't create content.xml")
```

internal/view/js/page/home.js
```diff
@@ -26,6 +26,9 @@ var template = `
         <a title="Update archives" @click="showDialogUpdateCache(selection)">
             <i class="fas fa-fw fa-cloud-download-alt"></i>
         </a>
+        <a title="Download ebooks" @click="ebookGenerate(selection)">
+            <i class="fas fa-fw fa-book "></i>
```
fmartingr, line 30:
```suggestion
            <i class="fas fa-fw fa-book"></i>
```

internal/core/ebook.go
```diff
@@ -0,0 +1,238 @@
...
+	contentOpfWriter, err := epubWriter.Create("OEBPS/content.opf")
+	if err != nil {
+		return book, true, fmt.Errorf("can't create content.opf")
+	}
+	contentOpfWriter.Write([]byte(`<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" version="2.0" unique-identifier="BookId">
+  <metadata>
+    <dc:title>` + book.Title + `</dc:title>
+    <dc:language>en</dc:language>
+    <dc:identifier id="BookId">urn:uuid:12345678-1234-5678-1234-567812345678</dc:identifier>
```
fmartingr, line 97: Should we create an identifier for each bookmark? What does this imply?

internal/core/ebook.go
```diff
@@ -0,0 +1,238 @@
...
+	// get list of images tag in html
+	imageList, _ := getImages(book.HTML)
+	imgRegex := regexp.MustCompile(`<img.*?src="([^"]*)".*?>`)
+
+	// Create a set to store unique image URLs
+	imageSet := make(map[string]bool)
+
+	// Download image in html file and generate new html
+	html = book.HTML
+	for _, match := range imgRegex.FindAllStringSubmatch(book.HTML, -1) {
```
fmartingr, line 171: In the future, we should try to aim and use the content we already have stored, but that could be complicated and we need to migrate from warc to obelisk (and potentially allow other archival generators) so this is just a general comment and not that you need to perform any action, just so you have that in mind.

internal/webserver/handler-ui.go
```diff
@@ -332,3 +332,65 @@ func (h *handler) serveBookmarkArchive(w http.ResponseWriter, r *http.Request, p
...
+	// Get bookmark ID from URL
+	strID := ps.ByName("id")
+	id, err := strconv.Atoi(strID)
+	checkError(err)
+	// Get bookmark in database
+	bookmark, exist, err := h.DB.GetBookmark(ctx, id, "")
+	checkError(err)
+
+	if !exist {
+		panic(fmt.Errorf("bookmark not found"))
```
fmartingr, line 349: I know there are panics all around the API, but we should avoid it unless there's no other way. Just log an error here, return a 404 response and end the function.

internal/webserver/handler-ui.go
```diff
@@ -332,3 +332,65 @@ func (h *handler) serveBookmarkArchive(w http.ResponseWriter, r *http.Request, p
...
+		if err != nil {
+			newPath := path.Join(h.RootPath, "/login")
+			redirectURL := createRedirectURL(newPath, r.URL.String())
+			redirectPage(w, r, redirectURL)
+			return
+		}
+	}
+
+	// Check if it has ebook.
+	ebookPath := fp.Join(h.DataDir, "ebook", strID+".epub")
```
fmartingr, line 364: Can't you use `bookmark.HasEbook` here?

internal/view/js/page/home.js
```diff
@@ -583,6 +588,59 @@ export default {
...
+                if (!response.ok) throw response;
+                return response.json();
+            }).then(json => {
+                this.selection = [];
+                this.editMode = false;
+                json.forEach(book => {
+                    // download ebooks
+                    const id = book.id;
+                    const ebook_url = new URL(`bookmark/${id}/ebook`, document.baseURI);
+                    const downloadLink = document.createElement("a");
```
fmartingr, line 625: Just curious: Why did you use this method and not a `location.href` ?

internal/view/js/component/bookmark.js
```diff
@@ -112,6 +126,17 @@ export default {
...
+		},
+		downloadebook() {
+            const id = this.id;
+            const ebook_url = new URL(`bookmark/${id}/ebook`, document.baseURI);
+            const downloadLink = document.createElement("a");
+            downloadLink.href = ebook_url.toString();
+            downloadLink.download = `${this.title}.epub`;
+            downloadLink.click();
+		},
+		generateEbook() {
```
fmartingr, line 138: How do I trigger this from the UI?

fmartingr (changes requested):
Hey, I added a few comments around the code and also have some general questions/comments:

- Please try to use `errors.Wrap(err, "")` when possible, so we start properly propagating errors around and improve debugging.
- Related to avoid, please avoid `panic`. If you don't have access to a proper logger just use `log.Println("error: XXX")` for now.
- Where are you trying the resulting ebooks? I tried Apple Books and it doesn't import (nor it gives any information why, because Apple). Tried out https://epub-reader.online/ and it opens them but no images are shown (but it could just be the site because using `unzip -l` over the generated epub I see the image there).


Thanks a lot for the huge effort in this, is a really good feature to have! And also thank you for leaving the API as clean as a possible. It will make my work under #497 easier to rebase once this lands on `master`.

Once you address your final tasks and the linter errors I think this is good to go and we can merge into master and start dogfooding it. :)

internal/webserver/handler-ui.go
fmartingr: That's legacy code I had no involvement with, I assume this was just done as the first draft for shiori and just never got updated. It will get fixed in #497

internal/view/js/page/home.js
fmartingr: Oh! That's nice :)

fmartingr:
> > Where are you trying the resulting ebooks? I tried Apple Books and it doesn't import (nor it gives any information why, because Apple). Tried out https://epub-reader.online/ and it opens them but no images are shown (but it could just be the site because using unzip -l over the generated epub I see the image there).
> 
> I test the result ebook in calibre, zathura, kindle (with KOReader) and work fine in all of them (image too). Can you give me more details? For example, send me a sample EPUB file that work on apple books?

That's just fine, I was just curious. You tested it in enough platforms. The only books I have on apple books are bought so I can't share them around. I will try to compare one or check around, is not something I use often so it's not a problem for me to be honest.

fmartingr:
> > Where are you trying the resulting ebooks? I tried Apple Books and it doesn't import (nor it gives any information why, because Apple). Tried out https://epub-reader.online/ and it opens them but no images are shown (but it could just be the site because using unzip -l over the generated epub I see the image there).
> 
> I test the result ebook in calibre, zathura, kindle (with KOReader) and work fine in all of them (image too). Can you give me more details? For example, send me a sample EPUB file that work on apple books?

I have one public domain, the UnCurled book from Curl: https://daniel.haxx.se/uncurled/

internal/webserver/handler-ui.go
fmartingr: The easiest way is use `http.NotFound(r, w)`. #497 is a huge API refactor so don't worry about that.

internal/core/ebook.go
fmartingr: If the field is not required or we don't know the usage, we can avoid it for now. Keep me posted 👍

internal/webserver/handler-ui.go
fmartingr: Understood, no problem. We can resolve that at a later date.

poire-z:
> * better CSS for ebook. i pick up css from [here](https://github.com/poire-z/crengine/blob/2b7a96c2778291d2b136abbff60715a5aa0e0925/cr3gui/data/epub.css) with very small change. I think it should be tested in long term to determine embedded CSS is good enough or not. Thanks @poire-z. @poire-z if you don't like we use this CSS here [519e10d](https://github.com/go-shiori/shiori/commit/519e10d6cebe24206ce72d6f843c37a8d4ac1915), please comment me.

No problem, pick what you want :)
But, I'm quite sure you shouldn't pick all that (ie. there's tons of crengine specific stuff that won't work elsewhere, and that you don't need even if you're going to read these EPUBs with KOReader).

Any EPUB reading app will have its own "user agent" stylesheet. If it uses some web browser engine, it will have something ressembling more what you get with KOReader's html5.css.
koreader/crengine epub.css or html5.css are 2 alternative KOReader/crengine "user agent" stylesheets: ie. `b, strong { font-weight: bold; }` is what tells crengine that content of `<b>...</b>` should use a bold font.
ALL/ANY EPUB reading app will know how to do that, so you don't need to add it in your EPUB embedded CSS. 99% of what is in epub.css or html5.css is like that, stuff reading apps and web browsers know about. No need to repeat it in each EPUB you generate :)

Mostly the only stuff we do in epub.css that browsers don't do (and that html5.css doesn't do either, to mimic web browsers) is the handling of paragraph `<p>` : we don't want top/bottom margins, and we want `text-indent: 1.2em`, so it looks like a real book and not like a web page.
So, your embedded CSS could/should just pick these little tweaks, to adjust the default behaviour of html5.css and web browsers.
But this is also a question of taste :) Some of your users may prefer the web page look, others the book look, and may be some of that can be tweaked on the apps (like with KOReader's style tweaks).
So, it's a choice people of this project will have to do :)

internal/webserver/handler-api.go
```diff
@@ -458,6 +465,103 @@ func (h *handler) apiUpdateBookmark(w http.ResponseWriter, r *http.Request, ps h
...
+	if len(bookmarks) == 0 {
+		panic(errors.Wrap(err, "no bookmark with matching ids"))
+	}
+
+	// Fetch data from internet
+	mx := sync.RWMutex{}
+	wg := sync.WaitGroup{}
+	chDone := make(chan struct{})
+	chProblem := make(chan int, 10)
+	semaphore := make(chan struct{}, 10)
```
fmartingr, line 500: I'm assuming this semaphore is to limit concurrent download to this channel size, right?

internal/core/processing.go
```diff
@@ -125,6 +126,22 @@ func ProcessBookmark(req ProcessRequest) (book model.Bookmark, isFatalErr bool,
 		}
 	}
 
+	// If needed, create ebook as well
+	if book.CreateEbook {
+		ebookPath := fp.Join(req.DataDir, "ebook", fmt.Sprintf("%d.epub", book.ID))
```
fmartingr, line 131: We will need to be cautious in the future and do not remove content, archive or bookmark if we find problems when updating the cache. We could save the new one to a temporary location while we work, and if everything succeeds, replace the current one. We can create an issue for this.

internal/core/ebook.go
```diff
@@ -0,0 +1,252 @@
...
+	fp "path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/go-shiori/shiori/internal/model"
+	"github.com/pkg/errors"
+)
+
+func EbookGenerate(req ProcessRequest) (book model.Bookmark, isFatalErr bool, err error) {
```
fmartingr, line 19:
```suggestion
func GenerateEbook(req ProcessRequest) (book model.Bookmark, isFatalErr bool, err error) {
```

fmartingr: Do we need `isFatalErr` ?

internal/core/ebook.go
```diff
@@ -0,0 +1,252 @@
...
+		return book, true, errors.Wrap(err, "can't write into mimetype file")
+	}
+
+	// Create the container.xml file
+	containerWriter, err := epubWriter.Create("META-INF/container.xml")
+	if err != nil {
+		return book, true, errors.Wrap(err, "can't create container.xml")
+	}
+
+	_, err = containerWriter.Write([]byte(`<?xml version="1.0" encoding="UTF-8"?>
```
fmartingr, line 89: Create issue for: Move ebook content to constant files or assets using embed.

fmartingr (changes requested):
Outstanding work @Monirzadeh ! I'd love to see this merged ASAP, I've just made some small comments around naming. If you could also some tests to the ebook generation so we have at least that part covered (don't bother with the API now) this will be good to go.

I only had some spare time to check the code and I have not tested running it locally, but will do it later, though it seems it would work as it did on my previous test.

Amazing work! 👏

fmartingr:
> @poire-z: No problem, pick what you want :)

My only concert about this is licensing, since [crengine](https://github.com/poire-z/crengine) is GPLv2 and Shiori is MIT, according to the GPLv2 licence we should distribute it under the same license.

I'm sorry I'm not very well versed in this kind of things, I will need to research this further before committing that change.

poire-z:
> My only concert about this is licensing, since [crengine](https://github.com/poire-z/crengine) is GPLv2 and Shiori is MIT, according to the GPLv2 licence we should distribute it under the same license.
I'm sorry I'm not very well versed in this kind of things, I will need to research this further before committing that change.

^ no need, you don't need to pick anything from koreader/crengine, and @Monirzadeh has removed from this PR the (uneeded and complex) stylesheet he picked from us.
If you need to later add stuff to your EPUB stylesheet, you'll just add bits, and there's no owner/license/copyright on using `P { margin: 0; text-indent: 1.2em }` :)

fmartingr:
> > If you could also some tests to the ebook generation so we have at least that part covered (don't bother with the API now) this will be good to go.
> 
> Any specific testing library?

Golang's builtin `testing` library is just fine.

fmartingr:
> ^ no need, you don't need to pick anything from koreader/crengine, and @Monirzadeh has removed from this PR the (uneeded and complex) stylesheet he picked from us.
> If you need to later add stuff to your EPUB stylesheet, you'll just add bits, and there's no owner/license/copyright on using `P { margin: 0; text-indent: 1.2em }` :)

Awesome. Thank you @poire-z !

internal/core/ebook.go
fmartingr: I meant moving the XML content to constants somewhere else (at the top of the file?) instead of having them inline in the code's logic.

internal/core/processing.go
fmartingr: Yeah sure, it was just a comment so we don't forget this in the future, and we can fix everything at once.

internal/core/ebook.go
fmartingr: Ideally fatal errors only happen at startup, if we are doing this per-request fatal errors make no sense, because they will stop the server. We will log the error and return an appropriate http status code to the user depending on the error. Thanks for the changes! All of that will go away with the new API.

internal/core/ebook_test.go
```diff
@@ -0,0 +1,328 @@
...
+	if info.Mode().Perm() != 0444 {
+		t.Errorf("ebook directory has incorrect permissions: expected 0444, got %o", info.Mode().Perm())
+	}
+	assert.False(t, bookmark.HasEbook)
+}
+
+// Add more unit tests for other scenarios that missing specialy
+// can't create ebook directory and can't write situatuin
+// writing inside zip file
+// html variable that not export and image download loop
```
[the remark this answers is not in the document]
fmartingr: Maybe that would mean a refactor of the logic function to allow easy testing. Some of that stuff could be tested when the epub is generated directly from the archive file. Don't try to cover 100% of the cases just yet, just the core logic is enough for now. When people start using it more issues may arise and we can refactor and create better tests that way.

internal/core/ebook_test.go
```diff
@@ -0,0 +1,328 @@
...
+	assert.NoError(t, err)
+}
+
+func TestGenerateEbook_InvalidBookmarkID_ReturnsError(t *testing.T) {
+	tempDir := t.TempDir()
+	defer os.RemoveAll(tempDir)
+	mockRequest := core.ProcessRequest{
+		Bookmark: model.Bookmark{
+			ID:            0,
+			URL:           "",
```
fmartingr, line 43: Do we need to manually set all object attributes for this test?

fmartingr (approved):
Awesome work @Monirzadeh, thanks a lot for all your work an involvement on this!

internal/core/ebook_test.go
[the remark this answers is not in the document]
fmartingr: You mean the UI not updating when we save a bookmark?

PR #627, opened 2023-07-01
thumb image will be not accesable anymore if it is private fix #578

### LINKED ISSUE #578: [Security] Thumbnails of private bookmarks are accessible without authentication
## Data
- **Shiori version**:  v1.5.4 (latest release)
- **Database Engine**: SQLite
- **Operating system**: `Raspbian GNU/Linux 11 (bullseye)`, `Linux raspberrypi 5.15.84-v7+ #1613 SMP Thu Jan 5 11:59:48 GMT 2023 armv7l GNU/Linux`
- **CLI/Web interface/Web Extension**: Web server (affects the server itself, not the web UI)

If the `/bookmark/{ID}/thumb` HTTP endpoint of a Shiori instance running in `serve` mode is reached, where ID is the numeric id of an existing saved bookmark, the image file that is a thumbnail of the corresponding bookmark is returned, **regardless of** if the bookmark in question is private **and** the current HTTP request is not authenticated.

## Expected behavior
If the above mentioned endpoint is reached, the software should return data of the thumbnail image file **only if at least one** of the following conditions are met, otherwise it should redirect to the login page exactly as it already correctly happens with the `/bookmark/{ID}/content` and `/bookmark/{ID}/archive` endpoints (ot, at the very least, deny access with an error):
- The bookmark is marked as public in the database
- The bookmark is marked as private in the database **but** the request is correctly authenticated

## To Reproduce
1. Know the ID of a private bookmark (trivial due to IDs not being random)
2. Verify that, without having a logged-in session in an account that owns the bookmark, it is not readable from either the `/bookmark/{ID}/content` or `/bookmark/{ID}/archive` HTTP endpoints of a Shiori instance running in `serve` mode (which is **correct behavior**)
3. Try to access the `/bookmark/{ID}/thumb` still without having a valid session, and notice that the thumbnail file is accessible as if the article it belongs to wasn't private (which **shouldn't happen**)

## Screenshots
Not applicable.

## Curl demonstration
```txt
$ curl --output - https://PRIVATE/bookmark/2/archive/
<a href="/login?dst=%2Fbookmark%2F2%2Farchive%2F">Moved Permanently</a>.

$ curl --output - https://PRIVATE/bookmark/2/thumb
{Binary data of the image file}
```

Monirzadeh:
Thanks, @andrigamerita
it checks acceptability of thumb before serve that to user
fix #578

### DIFF
--- a/internal/webserver/handler-ui.go
+++ b/internal/webserver/handler-ui.go
@@ -205,10 +205,33 @@ func (h *handler) serveBookmarkContent(w http.ResponseWriter, r *http.Request, p
 // serveThumbnailImage is handler for GET /bookmark/:id/thumb
 func (h *handler) serveThumbnailImage(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 	// Get bookmark ID from URL
-	id := ps.ByName("id")
+	ctx := r.Context()
 
+	// Get bookmark ID from URL
+	strID := ps.ByName("id")
+
+	// Get bookmark from database
+	id, err := strconv.Atoi(strID)
+	checkError(err)
+	bookmark, exist, err := h.DB.GetBookmark(ctx, id, "")
+	checkError(err)
+
+	if !exist {
+		panic(fmt.Errorf("bookmark not found"))
+	}
+
+	// If it's not public, make sure session still valid
+	if bookmark.Public != 1 {
+		err = h.validateSession(r)
+		if err != nil {
+			newPath := path.Join(h.RootPath, "/login")
+			redirectURL := createRedirectURL(newPath, r.URL.String())
+			redirectPage(w, r, redirectURL)
+			return
+		}
+	}
 	// Open image
-	imgPath := fp.Join(h.DataDir, "thumb", id)
+	imgPath := fp.Join(h.DataDir, "thumb", strID)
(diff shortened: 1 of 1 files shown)


internal/webserver/handler-ui.go
```diff
@@ -205,10 +205,33 @@ func (h *handler) serveBookmarkContent(w http.ResponseWriter, r *http.Request, p
...
+	strID := ps.ByName("id")
+
+	// Get bookmark from database
+	id, err := strconv.Atoi(strID)
+	checkError(err)
+	bookmark, exist, err := h.DB.GetBookmark(ctx, id, "")
+	checkError(err)
+
+	if !exist {
+		panic(fmt.Errorf("bookmark not found"))
```
fmartingr, line 220:
```suggestion
		log.Println("error: bookmark not found")
		return
```

fmartingr (changes requested):
Just a small thing and we can merge this.

Thanks for all your work!

fmartingr:
> Done.
> you can merge this now. I wish shiori get more progress in long run.
> Thanks for your time.

Thank you for all the effort you are putting into this :)

fmartingr (approved):

PR #639, opened 2023-07-09
feat: allow per-user settings and store them in database

Monirzadeh:
First step for real separate user in Shiori.
The goal is to move each user settings to the database and call each setting separately for each user.
I'm done some part of code but not push yet (it is not clean). It will be completed step by step until be ready for merge.
# Why we need this?
- I explain some part of reason in this comment https://github.com/go-shiori/shiori/discussions/338#discussioncomment-6376681.
- Maybe we can use this feature for cleaner code for #338 and make it per user 
- in the long term, maybe we can separate bookmark databases for each user. (we should keep public bookmark feature)

# TODO:

- [x] add needed field to the database
- [x] update model
- [x] update database.go to get database account information
- [x] update function for call data from database (sqlite)
- [x] settings of each user based on data called from database for that user
- [x] change logic to save configure just as on text field in SQLite
- [x] API for update setting record in database (better name for endpoint)
- [x] save settings from UI (do i need loading indicator while save in database?)
- [x] reflect updated value to the session
- [x] mechanism for check if configures is a valid json before add to database.
- [x] update function for call data from database (mysql)
- [x] update function for call data from database (pg)
- [x] add function saveSettings for other database. 
    - [x] MySQL 
    - [x]  PostgreSQL 
- [x] `SaveAccount` to `SaveAccountSettings` and `configures` to `config`
- [x] mechanism to migrate from old database to newer version and not break current instant of Shiori with data.
- [x] use json instead of text for config in database
- [x] Visitor account can't save settings because some settings not available for them. and validsetion check owner status fix that
- [x] can we use json or text for all database? Done all of them the same way in all database is better solution if it is possible
- [x] move update config in account variable to a function for more readable code
- [x] Done remaining TODO in code
- [x] down migrate script
- [x] fix some typo and update createEbook
- [x] Remove unneeded part because of `driver.valuer`
- [x] remove default config value for new user.
- [x] merge with master and solve conflicts
- [x] move to new API
- [x] solve validationsession 
- [x] remove jsonify
- [x] unit test for database
- [x] unit test for API

## Repeat for other database

# What I need to know?
- Specific security concern or test that should be pass?
- Any specific settings should be active for new user by default?
- ~~Should I change current API for update settings or create new one?~~

# Things make work unnecessary harder
- ~~current first user logic~~
- ~~first user (shiori with gopher) have problem before save that in database if user save new settings.  _Kinda bug in current logic_ be sure it solved.~~ after version 1.6 it not a problem anymore.

# Should be checked before merge:
- [x] does it work on SQLite?
- [x] does it work on MySQL?
- [x] does it work on PostgreSQL?
- [x] can migrate database without any problem?
- [ ] any specific security concern?


fmartingr:
Hey @Monirzadeh, thanks for looking into this! One side note though: Keep in mind that I will need to take over all API/DB modifications over to #497, so the more concise they are the better.

I will try to merge the partial effort ASAP so we can iterate further in that direction, but the fact that the two APIs have to coexist for a little bit will make things difficult. I will gladly listen to some feedback on that end.

internal/database/migrations/mysql/0001_initial.up.sql
```diff
@@ -1,8 +1,9 @@
 CREATE TABLE IF NOT EXISTS account(
-		id       INT(11)      NOT NULL AUTO_INCREMENT,
-		username VARCHAR(250) NOT NULL,
-		password BINARY(80)   NOT NULL,
-		owner    TINYINT(1)   NOT NULL DEFAULT '0',
+		id       INT(11)        NOT NULL AUTO_INCREMENT,
+		username VARCHAR(250)   NOT NULL,
+		password BINARY(80)     NOT NULL,
+		owner    TINYINT(1)     NOT NULL DEFAULT '0',
+		configures VARCHAR(500) NOT NULL,
```
fmartingr, line 6: I'd go with `config` or `configuration` as the name. And I'm not sure if you are using this as test, but the table alteration should be on a new migration (a new file). You need to use `ALTER TABLE` to modify the table, to allow current shiori deployments to upgrade seamlessly to this new schema.

Just add a new file adding 1 to the number of the previous one (if the last one is `0004`, you create `0005` and so on), and put the `ALTER TABLE` inside, it should work automatically if the SQL is right.

fmartingr: Also, I think we can use `JSON` in all engines, I'm creating a snippet for you to check, because we should be able to use sqlx to deserialize directly to a struct.

fmartingr: Here's a playground: https://go.dev/play/p/QDlGfg2TMYZ

It will probably timeout due to the dependency download, but it should work locally so you can get an idea. it should be even easier for psql and mariadb I guess.

internal/database/database.go
```diff
@@ -66,6 +75,9 @@ type DB interface {
 	// SaveAccount saves new account in database
 	SaveAccount(ctx context.Context, a model.Account) error
 
+	// SaveSettings saves settings for specific user in database
+	SaveSettings(ctx context.Context, a model.Account) error
```
fmartingr, line 71: I'd use `SaveAccountSettings` for simplicity. I'm still not sure if we actually need a new method for this or if we could use the already existing `SaveAccount`.

internal/model/model.go
```diff
@@ -30,8 +30,9 @@ type Bookmark struct {
...
 type Account struct {
-	ID       int    `db:"id"       json:"id"`
-	Username string `db:"username" json:"username"`
-	Password string `db:"password" json:"password,omitempty"`
-	Owner    bool   `db:"owner"    json:"owner"`
+	ID       int    `db:"id"                   json:"id"`
+	Username string `db:"username"             json:"username"`
+	Password string `db:"password"             json:"password,omitempty"`
+	Owner    bool   `db:"owner"                json:"owner"`
+	Config   string `db:"config"               json:"config"`
```
fmartingr, line 37: Have you checked https://github.com/go-shiori/shiori/pull/639#discussion_r1265419217 ?

fmartingr (commented):
Left some comments

internal/model/model.go
fmartingr: In your latest approach, from what I can see you can remove the empty slice check (since `for rows.Next()` wont do anything if the result set is empty anyway).

Also, if you use `Queryx` instead of `Query`, you can use `rows.StructScan` later in the for loop, which will help using a struct for the configuration as well.

Does this help?

internal/database/database.go
```diff
@@ -108,3 +125,40 @@ func (db *dbbase) withTx(ctx context.Context, fn func(tx *sqlx.Tx) error) error
...
+}
+
+func (c *UserConfig) Scan(value interface{}) error {
+	b, ok := value.([]byte)
+	if !ok {
+		return fmt.Errorf("unexpected type for UserConfig: %T", value)
+	}
+
+	return json.Unmarshal(b, c)
+}
```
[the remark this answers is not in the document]
fmartingr: Why can't you have this logic in the models?

internal/database/database.go
```diff
@@ -108,3 +125,40 @@ func (db *dbbase) withTx(ctx context.Context, fn func(tx *sqlx.Tx) error) error
...
+	var jsonData interface{}
+	err := json.Unmarshal([]byte(input), &jsonData)
+	if err != nil {
+		return err
+	} else {
+		return nil
+	}
+}
+
+func Jsonify(uc model.UserConfig) ([]byte, error) {
```
fmartingr, line 140: A simple `json.Marshal` of `model.UserConfig` should be enough to serialize it. This logic belong outside of the database. Did you encounter any issues with this?

internal/database/database.go
```diff
@@ -108,3 +125,40 @@ func (db *dbbase) withTx(ctx context.Context, fn func(tx *sqlx.Tx) error) error
 
 	return err
 }
+
+// get a string and return errors if it is not in json format
+func IsJson(input string) error {
```
fmartingr, line 130: I believe you can remove this function.

internal/model/model.go
```diff
@@ -30,8 +35,29 @@ type Bookmark struct {
...
-	Password string `db:"password" json:"password,omitempty"`
-	Owner    bool   `db:"owner"    json:"owner"`
+	ID       int        `db:"id"                   json:"id"`
+	Username string     `db:"username"             json:"username"`
+	Password string     `db:"password"             json:"password,omitempty"`
+	Owner    bool       `db:"owner"                json:"owner"`
+	Config   UserConfig `db:"config"               json:"config"`
+}
+
+type UserConfig struct {
```
fmartingr, line 45: If you implement [`driver.Valuer`](https://pkg.go.dev/database/sql/driver#Valuer) here, you can avoid (un)marshalling in the database logic:

```
func (c UserConfig) Value() (driver.Value, error) {
	return json.Marshal(c)
}
```

I'm sorry about this, I checked the example I provided and it was wrong. I made it really quick and forget that I left the marshalling logic instead of properly impementing the `Value() (driver.Value, error)` method.

With this, you can do:

```
db.Exec("UPDATE account SET config = ? WHERE username = ?", account.Config, account.Username)
```

directly, without manually marshalling anything.

internal/database/mysql.go
```diff
@@ -559,11 +574,28 @@ func (db *MySQLDatabase) GetAccounts(ctx context.Context, opts GetAccountsOption
 	query += ` ORDER BY username`
 
 	// Fetch list account
-	accounts := []model.Account{}
-	err := db.SelectContext(ctx, &accounts, query, args...)
-	if err != nil && err != sql.ErrNoRows {
+	rows, err := db.Queryx(query, args...)
```
fmartingr, line 577: You should be able to use the same `SelectContext` here since `model.Account` implements [`driver.Scanner`](https://pkg.go.dev/database/sql#Scanner). Did you encounter any issues?

My guess is the same for all database engines.

internal/database/mysql.go
```diff
@@ -572,12 +604,17 @@ func (db *MySQLDatabase) GetAccounts(ctx context.Context, opts GetAccountsOption
 // Returns the account and boolean whether it's exist or not.
 func (db *MySQLDatabase) GetAccount(ctx context.Context, username string) (model.Account, bool, error) {
 	account := model.Account{}
-	if err := db.GetContext(ctx, &account, `SELECT
-		id, username, password, owner FROM account WHERE username = ?`,
-		username,
-	); err != nil {
-		return account, false, errors.WithStack(err)
-	}
+	row := db.QueryRowx(`SELECT
```
fmartingr, line 607: You should be able to use the same `GetContext` here since `model.Account` implements [`driver.Scanner`](https://pkg.go.dev/database/sql#Scanner). Did you encounter any issues?

My guess is the same for all database engines.

internal/webserver/handler-api.go
```diff
@@ -108,6 +108,16 @@ func (h *handler) apiLogin(w http.ResponseWriter, r *http.Request, ps httprouter
 		genSession(model.Account{
```
fmartingr, line 108: This endpoint is no longer used in latest `master` so you can ignore it. Merge master into you branch to see latest changes.

internal/webserver/handler-api.go
```diff
@@ -846,6 +856,41 @@ func (h *handler) apiUpdateAccount(w http.ResponseWriter, r *http.Request, ps ht
 	fmt.Fprint(w, 1)
 }
 
+// apiUpdateSettings is handler for PUT /api/accounts
+func (h *handler) apiUpdateSettings(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
```
fmartingr, line 860: Are you up for moving this endpoint to the new API under `PATCH /api/v1/auth/account`? That endpoint will be for the current logged in user, so there's no custom validation required.

internal/webserver/handler.go
```diff
@@ -129,3 +129,19 @@ func (h *handler) validateSession(r *http.Request) error {
 
 	return nil
 }
+
+// validateSession checks whether user session is still valid or not without owner status
+func (h *handler) validateSessionWithoutOwnerStatus(r *http.Request) error {
```
fmartingr, line 134: See my comment in `handler-api.go`.

internal/webserver/server.go
```diff
@@ -202,6 +202,7 @@ func ServeApp(cfg Config) error {
 
 	router.GET(jp("/api/accounts"), withLogging(hdl.apiGetAccounts))
 	router.PUT(jp("/api/accounts"), withLogging(hdl.apiUpdateAccount))
+	router.PUT(jp("/api/accountssettings"), withLogging(hdl.apiUpdateSettings))
```
fmartingr, line 205: See my comment in `handler-api.go`

internal/view/js/page/base.js
```diff
@@ -14,16 +14,16 @@ export default {
...
+					ListMode: false,
+					NightMode: false,
+					HideThumbnail: false,
+					HideExcerpt: false,
 
-					keepMetadata: false,
-					useArchive: false,
+					KeepMetadata: false,
+					UseArchive: false,
 					createEbook: false,
```
fmartingr, line 25:
```suggestion
					CreateEbook: false,
```

internal/view/js/page/base.js
```diff
@@ -14,16 +14,16 @@ export default {
...
-					showId: false,
-					listMode: false,
-					nightMode: false,
-					hideThumbnail: false,
-					hideExcerpt: false,
+					ShowId: false,
+					ListMode: false,
+					NightMode: false,
+					HideThumbnail: false,
+					HideExcerpt: false,
```
fmartingr, line 22: Remove empty line.

internal/view/js/page/base.js
fmartingr: Aren't we storing this in the database?

internal/view/js/page/setting.js
```diff
@@ -155,11 +178,23 @@ export default {
 						this.showErrorDialog("Password does not match");
 						return;
 					}
+                    const defaultconfig = {
```
fmartingr, line 181: Is there a way we won't provide a default configuration directly from the backend? With `StructScan`, even if you have `{}` in the database, we will return the empty struct as default (everything to false.)

internal/view/login.html
```diff
@@ -109,6 +109,7 @@
 
 						// Save account data
 						localStorage.setItem("shiori-account", JSON.stringify(json.account));
+                        localStorage.setItem("shiori-setting", JSON.stringify(json.account.config));
```
fmartingr, line 112: Why are you not getting the config from the `shiori-account` storage? It's getting stored there as well.

internal/model/model.go
```diff
@@ -30,8 +35,29 @@ type Bookmark struct {
...
+	HideThumbnail bool `json:"HideThumbnail"`
+	HideExcerpt   bool `json:"HideExcerpt"`
+	NightMode     bool `json:"NightMode"`
+	KeepMetadata  bool `json:"KeepMetadata"`
+	UseArchive    bool `json:"UseArchive"`
+	MakePublic    bool `json:"MakePublic"`
+}
+
+func (c *UserConfig) Scan(value interface{}) error {
+	b, ok := value.([]byte)
```
fmartingr, line 57: I believe handling `[]byte` is enough here if we use `JSON` types for all engines, but if for any reason an engine uses `TEXT/VARCHAR` datatype, we should handle `string` as well.

fmartingr (changes requested):
Outstanding work @Monirzadeh! 

First of all, apologies for the confusion on the database custom struct types, the example I provided was wrong and you can simplify the code by implementing the [`driver.Valuer`](https://pkg.go.dev/database/sql/driver#Valuer) interface. I have left some example code in comments below.

Apart from that, there's only minor things going around that I have left in comments.

Please, remember to merge latest `master` into your branch so you can get the latest changes, and if you are up for it, you can move the API logic to the new code under `internal/http`. I hope you find it easy to use. 

Remember to add some unit tests (specially for the database logic) so it runs automatically on all engines. You have some under [`internal/database/database_test.go`](https://github.com/go-shiori/shiori/blob/master/internal/database/database_test.go).

internal/http/routes/api/v1/auth.go
```diff
@@ -23,6 +23,7 @@ func (r *AuthAPIRoutes) Setup(group *gin.RouterGroup) model.Routes {
 	group.GET("/me", r.meHandler)
 	group.POST("/login", r.loginHandler)
 	group.POST("/refresh", r.refreshHandler)
+	group.PUT("/account", r.settingsHandler)
```
fmartingr, line 26:
```suggestion
	group.PATCH("/account", r.settingsHandler)
```

internal/http/routes/api/v1/auth.go
```diff
@@ -150,6 +163,40 @@ func (r *AuthAPIRoutes) meHandler(c *gin.Context) {
...
+
+	if err := payload.IsValid(); err != nil {
+		response.SendError(c, http.StatusBadRequest, err.Error())
+		return
+	}
+	ctx := context.NewContextFromGin(c)
+	if ctx.UserIsLogged() {
+		response.SendError(c, http.StatusForbidden, nil)
+		return
+	}
```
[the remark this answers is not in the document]
fmartingr: You are not returning anything, really. Just finishing the function early, since this functions doesn't have return values :)

You can move this to the beginning of the function, too. Since you don't need to do anything if the user is not logged in.

internal/database/migrations/postgres/0002_config.down.sql
fmartingr, line 1: Up and down migrations should have the same number (0002_config.up.sql, 0002_config.down.sql). Up migration is when the user migrates forward (updates a version) and the down migration is required if a rollback is performed, to leave the database in the previous state.

fmartingr:
> Can you give me a unit test example for a new API? I don't know how should I first log in than change settings with API in unit test.

You can check the refresh token test: https://github.com/go-shiori/shiori/blob/master/internal/http/routes/api/v1/auth_test.go#L176-L182

It creates a valid token and makes a request with it. In your case you would need to create the account directly in the database with `dependencies.Database.SaveAccount`, in order to properly check the endpoint. I think the only endpoint that has tests that modify the database are from tags, they _should_ work, but let me know if you find any issues with that.

internal/http/routes/api/v1/auth.go
```diff
@@ -150,6 +163,40 @@ func (r *AuthAPIRoutes) meHandler(c *gin.Context) {
...
+		response.SendInternalServerError(c)
+		return
+	}
+
+	if err := payload.IsValid(); err != nil {
+		response.SendError(c, http.StatusBadRequest, err.Error())
+		return
+	}
+	ctx := context.NewContextFromGin(c)
+	if ctx.UserIsLogged() {
```
fmartingr, line 178:
```suggestion
	if !ctx.UserIsLogged() {
```

internal/http/routes/api/v1/auth.go
```diff
@@ -150,6 +163,40 @@ func (r *AuthAPIRoutes) meHandler(c *gin.Context) {
...
+		response.SendError(c, http.StatusForbidden, nil)
+		return
+	}
+
+	account, _, err := r.deps.Database.GetAccount(c, payload.Username)
+	if err != nil {
+		response.SendInternalServerError(c)
+		return
+	}
```
fmartingr, line 187:
```suggestion
	account := ctx.GetAccount()
```

internal/view/login.html
fmartingr: And why not update the settings stored in `shiori-account` when the user modifies them?

internal/database/database.go
fmartingr: I believe this function does not exist anymore, so I'm resolving this.

internal/http/routes/api/v1/auth.go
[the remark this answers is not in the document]
fmartingr: You need to send the new authentication JWT token with this request for this to work. The auth token is under `shiori-token` in the local storage. Send it as such: `Authorization: Bearer ${token}` and your request should work.

internal/http/routes/api/v1/auth.go
```diff
@@ -150,6 +163,34 @@ func (r *AuthAPIRoutes) meHandler(c *gin.Context) {
 	response.Send(c, http.StatusOK, ctx.GetAccount())
 }
 
+func (r *AuthAPIRoutes) settingsHandler(c *gin.Context) {
```
fmartingr, line 166: Remember to add the swagger tags to the handler so it will automatically show up in the [swagger documentation](https://github.com/go-shiori/shiori/blob/master/docs/APIv1.md).

fmartingr:
> I don't touch `shiori-token` but it not removes on logout from browser, is it normal?

You need to remove it on logout.

internal/view/assets/js/page/setting.js
```diff
@@ -105,13 +106,18 @@ export default {
...
+                Authorization: `Bearer ${authToken}`,
+                },
             }).then(response => {
                 if (!response.ok) throw response;
-                return response;
+                return response.json();
+            }).then(responseData => {
+                const responseString = JSON.stringify(responseData.message);
+                localStorage.setItem('shiori-account', responseString);
+                //return response.json();
```
fmartingr, line 120:
```suggestion
```

internal/http/routes/api/v1/auth_test.go
```diff
@@ -184,3 +185,51 @@ func TestRefreshHandler(t *testing.T) {
 		require.Equal(t, http.StatusAccepted, w.Code)
 	})
 }
+
+func TestSettingsHandler(t *testing.T) {
```
fmartingr, line 189: I would add a test here that checks that the user configuration is correctly updated in database as well.

internal/http/routes/api/v1/auth.go
```diff
@@ -163,27 +155,29 @@ func (r *AuthAPIRoutes) meHandler(c *gin.Context) {
 	response.Send(c, http.StatusOK, ctx.GetAccount())
 }
 
+// settingsHandler godoc
+// @Summary      Get config for the current logged in user and save that in database
```
fmartingr, line 159:
```suggestion
// @Summary      Perform actions on the currently logged-in user.
```

fmartingr (approved):
Amazing work @Monirzadeh ! I have left you some small comments but this seems good to go.

I'm going to hold this on until 1.6.1 first or until I check in which state we can ship 1.6.0 as per the roadmap.

Remember to update the branch from latest master too

fmartingr (approved):
Awesome, this is ready to go :)

fmartingr:
Hey @Monirzadeh let's fix the conflicts here and merge this directly for 1.6.0.

PR #646, opened 2023-07-15
fix download data image fix #645

### LINKED ISSUE #645: certain ebook exports are crashing shiori
## Data
- **Shiori version**: 1.5.5-rc.1
- **Database Engine**: mysql
- **Operating system**: docker on synology nas
- **CLI/Web interface/Web Extension**: web

I added an archive of a singlefile html file, but whenever shiori tries to create an ebook for it shiori crashes.

## Expected behavior
It should generate an ebook without crashing.

## To Reproduce
1. Add https://colorbash.neocities.org/archives/an-ode-to-jim-dandy-96aba193e3a0 to shiori
2. Click on 'Update Archive'
3. Check 'Update ebook as well'
4. Error ```<html> <head><title>502 Bad Gateway</title></head> <body> <center><h1>502 Bad Gateway</h1></center> <hr><center>openresty</center> </body> </html> (502)```
5. Shiori server crashes/restarts
6. The archive and the readable versions generate fine, but the ebook link downloads a 0 byte file

## Notes
This page is a singlefile html export, which i suspect might be part of the problem, I can test that out if it seems likely to be relevant. Using this as a workaround for websites that require sign-ins, since Shiori can't access those directly.

Monirzadeh:
This PR fix problem that download mechanism tried to get data/image.
It should not happen (they are not a real URL that point to the specific server, they are actual data)
by excluding them, we can handle data/image too.
Caliber or modern browser can handle this kind of links itself.
Will fix #645

for [Future reference](https://en.wikipedia.org/wiki/Data_URI_scheme)

### DIFF
--- a/internal/core/ebook.go
+++ b/internal/core/ebook.go
@@ -245,7 +245,9 @@ func GetImages(html string) (map[string]string, error) {
 	// Loop through all the matches and add them to the dictionary
 	for _, match := range imageTagMatches {
 		imageURL := match[1]
-		images[imageURL] = match[0]
+		if !strings.HasPrefix(imageURL, "data:image/") {
+			images[imageURL] = match[0]
+		}
 	}
 
 	return images, nil


fmartingr (approved):
Thanks for the quick fix!

Issue #634, opened 2023-07-07
Remove old cache if updating cache done without problem


Monirzadeh:
>We will need to be cautious in the future and do not remove content, archive or bookmark if we find problems when updating the cache. We could save the new one to a temporary location while we work, and if everything succeeds, replace the current one. We can create an issue for this.

_Originally posted by @fmartingr in https://github.com/go-shiori/shiori/pull/623#discussion_r1255429395_

PR #680, opened 2023-07-27
thumbnail will update in update archive

### LINKED ISSUE #524: The application does not remove the thumbnail file from the FS if it is removed in the original article
## Data
- **Shiori version**: commit [77269b7](https://github.com/go-shiori/shiori/commit/77269b7a6420220a0af1fc9c381b5319fbdde686)
- **Database Engine**: SQLite
- **Operating system**: Windows 8.1

The problem is described in the title of the issue.

## Expected behavior
If a thumbnail is removed from the original article, then it should also be removed from the file system.

## To Reproduce
1. Save any HTML page with thumbnail from the internet to the localhost and serve it with any way (e. g. `python -m SimpleHTTPServer 8080`.
2. Add a link to this page in Shiori. As you can see, an image file is created in the `thumb` directory.
3. Remove all text and all image tags from the saved page.
4. Click the update archive button. I also check both checkboxes: "Keep the old title and excerpt" and "Update archive as well".
5. As you can see, the site thumbnail disappeared from Shiori's interface, but it remained in the file system.

Monirzadeh:
This pull request update thumbnail after update archive version. fix #524
now bookmark thumbnail update based on last version of webpage.

### DIFF
--- a/internal/core/processing.go
+++ b/internal/core/processing.go
@@ -98,6 +98,12 @@ func ProcessBookmark(req ProcessRequest) (book model.Bookmark, isFatalErr bool, 
 			book.Title = book.URL
 		}
 
+		if book.CreateArchive {
+			archivePath := fp.Join(req.DataDir, "archive", fmt.Sprintf("%d", book.ID))
+			imgPath := fp.Join(req.DataDir, "thumb", fmt.Sprintf("%d", book.ID))
+			os.Remove(archivePath)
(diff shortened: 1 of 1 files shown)


fmartingr:
This is going to get tricky and will require a bigger refactor to make #524 and #634 work together, since on your changes you remove the current data without knowing if the new data will be processed correctly.

PR #679, opened 2023-07-27
refactor: Migrate ePub generation to go-epub

Monirzadeh:
This pull request solves the problem in #638 and #678
Please before merge test that in real device. Based on your report and in my debug system, it should work fine.
we move to `go-epub` in this pull request that result of that
fix #638 
fix #678 
fix #688
fix #635 

## TODO:
- [x] add `go-epub` to requirement
- [x] Support RTL
- [x] update HTML to use included image
- [x] fix no reference image file. duplicate file included
- [x] update unit test

PR #683, opened 2023-07-28
fix: Ensure bookmark files are correctly downloaded before deleting current ones

### LINKED ISSUE #524: The application does not remove the thumbnail file from the FS if it is removed in the original article

### LINKED ISSUE #634: Remove old cache if updating cache done without problem

Monirzadeh:
This pull request should recover old version of Archive and e-book if download Processes failed.
It should close #634 
fix #524
# TODO:
- [x] generate ebook to dstPath
- [x] Archive Recover
- [x] E-Book Recover
- [x] Thumb Recover
- [x] E-Book Download API should manage if ebook file exist not download as TODO in code
- [x] if download Failed and recover old file, it should report to user in UI
- [x] thumbnail will update in update archive #524 #680 
- [x] update old unit test.
- [x] add more unit test to cover `processing.go` 87.7% 
- [x] add test image file to repository
- [x] replace URL with sample file

# Things need more time:
should we recover thumb image? What if thumb image remove from source like #524. How we should detect that


fmartingr:
I'd say the path going forward should be: Download the bookmark archive/article/thumb/ebook to a temporary folder. If everything goes ok, we replace the old ones, if not, we raise an error. But it should be open enough to allow multiple operations.

This is a bigger task that you may think, and we need to refactor the logic into smaller chunks to made them easily testable.

internal/core/ebook.go
```diff
@@ -40,34 +40,22 @@ func GenerateEbook(req ProcessRequest) (book model.Bookmark, err error) {
...
-	if _, err := os.Stat(ebookDir); os.IsNotExist(err) {
-		err := os.MkdirAll(ebookDir, model.DataDirPerm)
-		if err != nil {
-			return book, errors.Wrap(err, "can't create ebook directory")
-		}
-	}
-	// create epub file
-	epubFile, err := os.Create(ebookfile)
+	// create temporary epub file
+	tmpFile, err := os.CreateTemp("", "ebook*.epub")
```
fmartingr, line 50: Why an asterisk in the file name?

internal/core/processing.go
```diff
@@ -236,12 +243,43 @@ func downloadBookImage(url, dstPath string) error {
...
 
+	err = MoveToDestination(dstPath, tmpFile)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func MoveToDestination(dstPath string, tmpFile *os.File) error {
```
fmartingr, line 261: I would use `MoveFileToDestination` in case we want to have something for folders later.
```suggestion
func MoveFileToDestination(dstPath string, tmpFile *os.File) error {
```

fmartingr: Add comment making clear that `dstPath` requires the filename. You can add a third argument for the filename, up to you.

internal/core/processing.go
```diff
@@ -101,6 +103,8 @@ func ProcessBookmark(req ProcessRequest) (book model.Bookmark, isFatalErr bool,
 		// Get image URL
 		if article.Image != "" {
 			imageURLs = append(imageURLs, article.Image)
+		} else {
+			os.Remove(imgPath)
```
fmartingr, line 107: I would remove this from here and let `downloadBookImage` do the cleanup work.

internal/core/processing.go
```diff
@@ -115,9 +119,6 @@ func ProcessBookmark(req ProcessRequest) (book model.Bookmark, isFatalErr bool,
 	}
 
 	// Save article image to local disk
-	strID := strconv.Itoa(book.ID)
-	imgPath := fp.Join(req.DataDir, "thumb", strID)
-
 	for _, imageURL := range imageURLs {
 		err = downloadBookImage(imageURL, imgPath)
```
fmartingr, line 123: Log something if `err != nil`

internal/view/assets/js/page/home.js
```diff
@@ -702,15 +702,35 @@ export default {
...
+                            if (data.createEbook && !book.hasEbook){
+                                faildCreateEbook.push(book.id);
+                            }
+                        }),
+
+                        this.showDialog({
+                            title: `Update Archive Error`,
+                            content: `Bookmarks Update Archive Faild : ${faildUpdateArchives.join(", ")} 
+                            Bookmarks  Create Ebook Faild: ${faildCreateEbook.join(",")} 
+                            We recovered the last available version.`,
```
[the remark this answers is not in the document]
fmartingr: Have you tried `<br>` ?

internal/webserver/handler-ui.go
```diff
@@ -320,7 +320,7 @@ func (h *Handler) ServeBookmarkEbook(w http.ResponseWriter, r *http.Request, ps
 	}
 
 	// Check if it has ebook.
-	ebookPath := fp.Join(h.DataDir, "ebook", strID+".epub")
+	ebookPath := fp.Join(h.DataDir, "ebook", strID)
```
fmartingr, line 323: Changing this will make it not backwards compatible. People that have ebooks stored and upgrade to this change wont be able to download their ebooks, ebook will be downloaded again and the old one (without extension) will be left in the filesystem.

internal/webserver/handler-api.go
```diff
@@ -110,7 +110,7 @@ func (h *Handler) ApiGetBookmarks(w http.ResponseWriter, r *http.Request, ps htt
 		strID := strconv.Itoa(bookmarks[i].ID)
 		imgPath := fp.Join(h.DataDir, "thumb", strID)
 		archivePath := fp.Join(h.DataDir, "archive", strID)
-		ebookPath := fp.Join(h.DataDir, "ebook", strID+".epub")
+		ebookPath := fp.Join(h.DataDir, "ebook", strID)
```
fmartingr, line 113: Read comment in `handler-ui.go`

internal/webserver/handler-api.go
```diff
@@ -285,7 +285,7 @@ func (h *Handler) ApiDeleteBookmark(w http.ResponseWriter, r *http.Request, ps h
 		strID := strconv.Itoa(id)
 		imgPath := fp.Join(h.DataDir, "thumb", strID)
 		archivePath := fp.Join(h.DataDir, "archive", strID)
-		ebookPath := fp.Join(h.DataDir, "ebook", strID+".epub")
+		ebookPath := fp.Join(h.DataDir, "ebook", strID)
```
fmartingr, line 288: Read comment in `handler-ui.go`

internal/webserver/handler-api.go
```diff
@@ -434,12 +434,32 @@ func (h *Handler) ApiDownloadEbook(w http.ResponseWriter, r *http.Request, ps ht
 				ContentType: contentType,
 			}
 
-			book, err = core.GenerateEbook(request)
-			content.Close()
+			// if file exist book return avilable file
+			strID := strconv.Itoa(book.ID)
+			ebookPath := fp.Join(request.DataDir, "ebook", strID)
```
fmartingr, line 439: Here you are not using the `${strID}.epub` filename. We need to be consistent. Maybe creating a function called `GetEbookFilename(book model.Bookmark) string` that receives the bookmark and returns the ebook filename?

internal/view/assets/js/page/home.js
```diff
@@ -702,15 +702,36 @@ export default {
...
+                                faildCreateEbook.push(book.id);
+                            }
+                        });
+                        if(faildCreateEbook.length > 0 || faildUpdateArchives.length > 0){
+                        this.showDialog({
+                            title: `Update Archive Error`,
+                            content: `Bookmarks Update Archive Faild : ${faildUpdateArchives.join(", ")} 
+                            Bookmarks  Create Ebook Faild: ${faildCreateEbook.join(",")} 
+                            We recovered the last available version.`,
+                            mainText: "OK",
```
fmartingr, line 725:
```suggestion
                            Bookmarks  Create Ebook Failed: ${faildCreateEbook.join(",")} 
                            Files that failed retrieval were not overwritten.`,
                            mainText: "OK",
```

fmartingr (changes requested):
I have done a first pass really quick. I'm concerned about the backwards compatibility if we change the ebook filename now for people that already have 1.5.5. 

I will need to make a second pass trying it out locally.

internal/core/ebook.go
```diff
@@ -223,6 +211,27 @@ img {
...
+	err = epubWriter.Close()
+	if err != nil {
+		return book, errors.Wrap(err, "failed to close EPUB writer")
+	}
+	err = tmpFile.Close()
+	if err != nil {
+		return book, errors.Wrap(err, "failed to close temporary EPUB file")
+	}
+	// open temporary file again
+	tmpFile, err = os.Open(tmpFile.Name())
```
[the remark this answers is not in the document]
fmartingr: Then why not let `MoveFileToDestination` handle opening/closing the file?

internal/core/processing.go
fmartingr: Oh, of course. You are right 👍

internal/view/assets/js/page/home.js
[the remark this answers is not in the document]
fmartingr: Maybe you need to edit the dialog too. Let's keep it all under one line for now, we can fix the texts later.

internal/webserver/handler-ui.go
fmartingr: We can do something, but it's pretty inconvenient and hard for distributors to build upon. 

Let's move the "ebook exists" logic to one function and let the function:
1. Check the filename without extension
    - If exists, `os.Rename` the file to `id.epub` 
2. Check the `id.epub` extension.

We can remove the first check after some versions.

internal/core/ebook.go
fmartingr: I was asking about the `*` symbol that was in the filename, I don't mind the extension.

internal/core/processing.go
```diff
@@ -115,26 +119,33 @@ func ProcessBookmark(req ProcessRequest) (book model.Bookmark, isFatalErr bool,
...
 	// Save article image to local disk
-	strID := strconv.Itoa(book.ID)
-	imgPath := fp.Join(req.DataDir, "thumb", strID)
-
-	for _, imageURL := range imageURLs {
-		err = downloadBookImage(imageURL, imgPath)
+	for i, imageURL := range imageURLs {
+		err = DownloadBookImage(imageURL, imgPath)
+		if err != nil {
+			if err.Error() == fmt.Sprintf("%s is not a supported image", imageURL) {
```
fmartingr, line 125: Instead of this, you can create a custom error and check if its in the chain with `errors.Is`:

```go
var ErrNoSupportedImageType = errors.New("unsupported image type")

// func() error
return ErrUnsupportedImageType

// usage

if err != nil && errors.Is(err, ErrUnsupportedImageType) {
    // ...
}
```

internal/core/processing_test.go
fmartingr, line 1: I would prefer if tests well contained to the logic they have, something like:

```
func TestMoveFileToDestination(t *testing.T) {
    t.Run("create dir", func(t *testing.T) {
        t.Run("fails", func(t *testing.T) {
            // test cae
        })
    })
}
```

You can even [create a custom struct to handle multiple inputs/outputs and just iteate over it preparing the test](https://github.com/mattermost/mattermost-plugin-mscalendar/blob/master/server/command/connect_test.go#L18-L86), executing the call and then checking the results.

internal/view/assets/js/page/home.js
```diff
@@ -702,15 +702,36 @@ export default {
...
-
-						json.forEach(book => {
-							var item = items.find(el => el.id === book.id);
-							this.bookmarks.splice(item.index, 1, book);
-						});
-					}).catch(err => {
-						this.selection = [];
-						this.editMode = false;
-						this.dialog.loading = false;
+
```
fmartingr, line 705: I think that to make this simpler for now (since we don't have a nice way of presenting this to the user) I would just expose if everything went right or not.

The error message could look something like `Not all bookmarks could have their contents updated, but no files were overwritten.`

In order to have some debug information, we could expose the IDs in a single `console.error()` statement along with a server log, maybe?

fmartingr (changes requested):
Hey @Monirzadeh, nice work on this. I have pointed out a few things, some minor.

One thing that still worries me is the migration from ebooks without extension to files with extension, you are not still handling that, right? We need to properly take care of that before moving this forward.

fmartingr:
> i decide not to do that in this PR so i revert that in [return .epub](https://github.com/go-shiori/shiori/pull/683/commits/89c04d69390d328fb22b1714dc580b8b027afe59) so it is work like as before.

Oh wait, so you wanted to change from `file.epub` to simply `file` (without extension)?. It thought it was the opposite! 

I think we should leave the extension, so there's nothing to change if that's the case.

internal/core/ebook.go
```diff
@@ -16,6 +16,10 @@ import (
 	"github.com/pkg/errors"
 )
 
+// this function get request and a destination path and will create an epub file in dstPath from that request
+// dstPath should be incouded file name with '.epub'
+// it will return a bookmark model and err
+// bookmark model later use for update UI shiori based on this function be sucssesful or not
```
fmartingr, line 22: Made some minor changes, since this should start with the function name.

I'm not sold on this function returning a `model.Bookmark`, but I don't want to change that for now.


```suggestion
// GenerateEbook receives a `ProcessRequest` and generates an ebook file in the destination path specified.
// The destination path `dstPath` should include file name with ".epub" extension
// The bookmark model will be used to update the UI based on whether this function is successful or not.
```

internal/core/processing.go
```diff
@@ -121,15 +123,14 @@ func ProcessBookmark(req ProcessRequest) (book model.Bookmark, isFatalErr bool,
...
-		if err != nil {
-			if err.Error() == fmt.Sprintf("%s is not a supported image", imageURL) {
-				log.Printf("Not found image for URL: %s", imageURL)
-				if i == len(imageURLs)-1 {
-					os.Remove(imgPath)
-				}
-			} else {
-				log.Printf("File download not successful for image URL: %s", imageURL)
+		if err != nil && errors.Is(err, ErrNoSupportedImageType) {
+			log.Printf("Not found image for URL: %s", imageURL)
```
fmartingr, line 127: Let's print the error with the URL for now, to be consistent.

```suggestion
			log.Printf("%s: %s", err.String(), imageURL)
```

fmartingr (approved):
Hey, amazing work on this! I have put out two small changes on a docstring and an error handler but this should be good to go as well.

I still think that `MoveFileToDestination` should receive two paths and it should handle opening the file by itself, but we can revisit that later on.

Do you need this merged to start working on the go-epub migration?

fmartingr:
> i should change things in #679 so i review `MoveFileToDestination` there
> i don't like `model.Bookmark` too but current logic for update UI need to return `model.Bookmark`

We can leave it like that for now, don't worry.

Issue #91, opened 2018-05-02
Anchors in URLs get encoded incorrectly


muesli:
When storing a URL with an anchor (e.g. https://foo#bar) the hash character gets encoded incorrectly as `%23`.

RadhiFadlillah:
@muesli thanks for the report. This happened because `url.ParseRequestURI` (which used for validating URL) always assume that the raw URL string doesn't have any `#fragment` suffix.

To fix this, I've reverted `ParseRequestURI` to normal `Parse` in [`1baf921`](https://github.com/RadhiFadlillah/shiori/commit/1baf921b491db8b1a2218c81cf32471bdd2d3fcd). I've also removed the fragment from URL in [`7532d36`](https://github.com/RadhiFadlillah/shiori/commit/7532d36e7e61ff60395f1819bd9010d7b723074e) because AFAIK the anchor tag (#) is only used by web browser, and not by web server.

RadhiFadlillah:
Hi @muesli, I’m closing this issue for now because :

- It has been fixed, and;
- It has been inactive for a while, so I guess this issue has been solved. 

Feels free to reopen it if needed.

Thank you.

mailinglists35:
this is not fixed. i still cannot add url with anchor. using latest release binary. if it is fixed in master, perhaps could you try make a new release or at least a snapshot of the current master?

mailinglists35:
i mean... i really want to store an url containing a fragment.... :) the fix was to remove support for fragments altogether :-(

PR #687, opened 2023-07-30
preserve fragment in URLs (#315)

### LINKED ISSUE #315: Feature request: preserve hash in URLs
I noticed that when creating bookmarks the hash and things after will all be dropped, which causes some troubles. since "#" is used by Mega Drive to identify keys or used for anchors in html so that we jump to a specific part of webpage.

anyway, any solutions? for now i have to paste the original url in excerpt after creating a bookmark, bit annoying😥

arakimo:
I encountered this issue myself while using shiori, so I figured I would have a go. I'm fairly new to git/programming, so... apologies in advance for any mistakes :)

Should fix #315

### DIFF
--- a/internal/core/url.go
+++ b/internal/core/url.go
@@ -52,15 +52,6 @@ func RemoveUTMParams(url string) (string, error) {
 		}
 	}
 
-	frag := tmp.Fragment
-	tmp.Fragment = ""
 	tmp.RawQuery = queryEncodeWithoutEmptyValues(queries)
-
-	// Append fragment before returning URL.
-	if frag == "" {
-		return tmp.String(), nil
-	} else {
-		returnUrl := tmp.String() + "#" + frag
-		return returnUrl, nil
-	}
+	return tmp.String(), nil
 }
--- a/internal/core/url.go
+++ b/internal/core/url.go
@@ -52,7 +52,11 @@ func RemoveUTMParams(url string) (string, error) {
 		}
 	}
 
+	frag := tmp.Fragment
 	tmp.Fragment = ""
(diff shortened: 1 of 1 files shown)


fmartingr:
Hey @arakimo, thanks for checking this out! 

Your solution works, but I'm just wondering if isn't it easier just to remove the `tmp.Fragment = ""` line. We are removing it manually and I'm not sure of the reason. Did you try that before?

arakimo:
I assumed that the fragment was being removed manually before `queryEncodeWithoutEmptyValues(queries)` to prevent the fragment from being encoded (i.e. turning the # into a %23), but I just tested your suggestion and it seems to work perfectly, so I'm not sure why `tmp.Fragment = ""` is there either.

It was mentioned in (#91) that the code used to use `ParseRequestURI()`, which unlike `Parse()` assumed that the string provided would not have a fragment. Maybe manually clearing the fragment is a remnant of that? I'm not sure.

fmartingr:
> I assumed that the fragment was being removed manually before `queryEncodeWithoutEmptyValues(queries)` to prevent the fragment from being encoded (i.e. turning the # into a %23), but I just tested your suggestion and it seems to work perfectly, so I'm not sure why `tmp.Fragment = ""` is there either.
> 
> It was mentioned in (#91) that the code used to use `ParseRequestURI()`, which unlike `Parse()` assumed that the string provided would not have a fragment. Maybe manually clearing the fragment is a remnant of that? I'm not sure.

Yeah, I think this wasn't working when using `url.ParseRequestURI`, but right now there are no calls to that method so this should work. Let's merge this in and test on development to see if there's anything wrong with it.

Thanks a lot for your work!

fmartingr (approved):

PR #691, opened 2023-08-06
docs: Updated compile/build information for macOS and Linux

z3cko:
Updated the information on how to build the source on Mac OS X and Linux.

### DIFF
--- a/internal/core/url.go
+++ b/internal/core/url.go
@@ -52,7 +52,6 @@ func RemoveUTMParams(url string) (string, error) {
 		}
 	}
 
-	tmp.Fragment = ""
 	tmp.RawQuery = queryEncodeWithoutEmptyValues(queries)
 	return tmp.String(), nil
 }


docs/Installation.md
```diff
@@ -30,10 +30,11 @@ On Windows, you can simply set the `PATH` by using the advanced system settings.
 
 ## Building From Source
 
-Shiori uses Go module so make sure you have version of `go >= 1.14.1` installed, then run:
+Shiori uses Go module so make sure you have version of `go >= 1.14.1` installed. In the cloned git repository run:
 
 ```
 go get -u -v github.com/go-shiori/shiori
```
fmartingr, line 36: I think it's better to just switch this command to `go install github.com/go-shiori/shiori`, right? That will build and install the command from source directly without cloning the repository (which is what the old documentaton wanted to achieve I guess).

What do you think?

PR #705, opened 2023-08-25
[docs update] Add guidance for deploying with Fly.io

jt-wang:
## what this PR is about
This PR updated `Installation.md`, to add guidance for deploying Shiori with Fly.io.

## why it's needed
As many people look for self-hosting bookmark service, it would be great if the docs can provide them with a straight-forward and easy way to quickly set up an instance without diving into too much tech details. 

## notes for maintainer
This change is only a doc update, and doesn't affect any functionality of the project.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -5,6 +5,7 @@ There are several installation methods available :
 - [Using Precompiled Binary](#using-precompiled-binary)
 - [Building From Source](#building-from-source)
 - [Using Docker Image](#using-docker-image)
+  - [Deploying with Fly.io](#deploying-with-flyio)
 
 <!-- /TOC -->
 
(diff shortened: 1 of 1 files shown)


fmartingr:
Hey @jt-wang , thanks for the update! Didn't knew it was that easy to work with fly.io.

One thing though. Since i'm currently reviewing and rewriting all documentation, would you mind adding this to the `fmartingr/issue663` branch? Just wanted to make sure the wording is ok for alternative ways of deployment and then push the documentation update all at once, which will trigger a new release. 

Thanks!

jt-wang:
@fmartingr Sure, thanks! Here's the new PR #706  pushed to `fmartingr/issue663`.

PR #706, opened 2023-08-25
[docs update] Add guidance for deploying with Fly.io

jt-wang:
## what this PR is about
This PR updated `Installation.md`, to add guidance for deploying Shiori with Fly.io.

## why it's needed
As many people look for self-hosting bookmark service, it would be great if the docs can provide them with a straight-forward and easy way to quickly set up an instance without diving into too much tech details. 

## notes for maintainer
This change is only a doc update, and doesn't affect any functionality of the project.
This PR has the same content as #705, but pushing to `fmartingr/issue663` per requested by @fmartingr .

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -48,6 +48,34 @@ docker pull ghcr.io/go-shiori/shiori
 
 > This docker image is based on the [Dockerfile.ci](../.github/workflows/docker/Dockerfile.ci)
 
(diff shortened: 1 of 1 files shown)


fmartingr:
Hey @jt-wang , thanks for the contribution! 

This will be live when we release 1.6.0. Thanks for making Shiori accessible to more folks :)

PR #710, opened 2023-08-31
set saner postgresql connection default and make user/password optional

Mic92:
This in particular fixes unix socket connections which requires no user/password to be set.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -45,12 +45,12 @@ Shiori uses an SQLite3 database stored in the above data directory by default. I
 
 ### PostgreSQL
 
-| Variable            | Description                                         |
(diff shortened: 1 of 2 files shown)


fmartingr:
Hey @Mic92 , are you using 1.5.5 or latest master? In the latest code we added a new `SHIORI_DATABASE_URL` that will replace the current variables in a future version and theoretically supports unix sockets. Would you mind giving that a try? I'm currently rewritting the docs to reflect all these changes and it will be public in 1.6.0.

Mic92:
Ok. I'll switch to that

fmartingr:
> Ok. I'll switch to that

If that doesn't work for you, feel free to submit a PR or open a request to allow sockets using the database uri.

PR #721, opened 2023-09-10
fix(db): handle usage of special characters in searches

### LINKED ISSUE #717: SQL injection?
## Data
- 1.5.5
- SQLite(???). I did nothing to setup 
- Devuan x64 
- Web interface 

Entering something like "hea;th" in search control results in "SQL logic error: fts5: syntax error near ";" (1) (500)"

## Expected behavior
More qualified error or something sane

## To Reproduce
See above

## Screenshots
NA

## Notes
NA

fmartingr:
The current queries are built by concatenating strings in the code, rather than in the database, meaning we lose sanitization on the database side.

This PR adds a new test that checks queries with some special characters to avoid failures, while doing:
- String concatenation for `LIKE` queries in the query directly
- Properly quoting FTS5 search queries in SQLite
- Properly parsing double quotes in SQLite queries for FTS5.

Fixes #717

### DIFF
--- a/internal/database/database_test.go
+++ b/internal/database/database_test.go
@@ -273,14 +273,21 @@ func testGetBookmarksWithSQLCharacters(t *testing.T, db DB) {
 	_, err := db.SaveBookmarks(ctx, true, book)
 	assert.NoError(t, err, "Save bookmarks must not fail")
 
-	characters := []string{";", "%", "_", "\\", "\""}
+	characters := []string{";", "%", "_", "\\", "\"", ":"}
 
 	for _, char := range characters {
-		t.Run(char, func(t *testing.T) {
+		t.Run("GetBookmarks/"+char, func(t *testing.T) {
 			_, err := db.GetBookmarks(ctx, GetBookmarksOptions{
 				Keyword: char,
 			})
 			assert.NoError(t, err, "Get bookmarks should not fail")
+		})
+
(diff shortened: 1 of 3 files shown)


fmartingr:
> is it possible to happen similar things (in another way for mysql) that we should handle that too?
> beside that everything is fine to me.

Tests run on all database backends, so at least the query does not err. I will check properly later when I address your suggestion 👍

fmartingr:
> i run locally this pull request but i get that error yet 🤔 (sqlite)

Yeah, I forgot about `GetBookmarksCount` 😓

fmartingr:
> everything work fine now just one minor problem
> 
> 1. add this page https://en.wikipedia.org/wiki/Percent_sign
> 2. search for this querry `"%"` it is not return any result
> 3. the page contain this text `"%" redirects here. For the mod sign as used in modular arithmetic`
> 
> you can't search for that special character. for example `0025"\/%` return you result because you have text `U+0025% PERCENT SIGN` or `025% PERCENT` and `025 PERCENT` both has result it ignore them completely. can we fix that?

Fixing that small use case would take a while, since there are several queries and the query is built without a query builder like `sqlx` or something like that, just appending SQL one on top of another, for three different engines.

is this a priority use case for you? It's not that I don't want it to work properly, but maybe the effort of fixing it just for some character is time that we can spend in other areas, and we can always reconsider and revisit it in the future.

PR #730, opened 2023-09-11
fix: properly parse mysql connection string, docs update

### LINKED ISSUE #729: Mysql connection string is not parsed properly
## Data
- **Shiori version**: v1.5.5
- **Database Engine**: mysql
- **Operating system**: linux
- **CLI/Web interface/Web Extension**: NA

You cannot provide a mysql connection string that is accepted by both the driver and the application.

The problem is that it parses the connection string looking for a schema, but it won't be stripped from the string that is passed to the driver:
https://github.com/go-shiori/shiori/blob/master/internal/database/database.go#L57

But the driver itself doesn't accept a scheme:
https://github.com/go-sql-driver/mysql#dsn-data-source-name

Furthermore, the new connection string is preferred, but the documentation is still using the old way.

## Expected behavior
It connects to mysql instance.

## To Reproduce
* Use a connection string without a scheme: `shiori:passwd@(mysql)/shiori?charset=utf8mb4`
* Use a connection string with a scheme: `mysql://shiori:passwd@(mysql)/shiori?charset=utf8mb4`

## Notes
I'll create a PR for it later on. :D

rutkai:
Mysql connection string update, docs update

Fixes #729

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
-# Exclude config file
+# Exclude IDE
 .vscode/
+.idea/
+
+# Exclude config file
 *.toml
 
 # Exclude executable file
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -16,18 +16,9 @@ services:
       - "postgres"
       - "mariadb"
     environment:
-      SHIORI_DBMS: mysql
       SHIORI_DIR: /srv/shiori
-      SHIORI_PG_USER: shiori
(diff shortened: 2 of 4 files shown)


fmartingr (approved):
Hey @rutkai, thanks for the finding and spending time fixing the code and documentation! I planned to review all documentation by 1.6.0, but it just too much and I would be holding things for quite a bit, so this is a good improvement, really appreciated :D

PR #732, opened 2023-09-14
feat: allow resize the dialogbox for bigger/hidpi screens

### LINKED ISSUE #731: "Edit bookmark" window is too small
## Data
- Shiori version 1.6.0-rc.1 (build 1d52569731e72d2dcb4236f3008912b58a310f6d) at 2023-09-11T17:23:40Z
- SQLite
- Devuan x64 
- Web interface 

"Edit bookmark" window is too small. On 4K screen it occupies too little space & can't be expanded.

## Expected behavior
Make window be a large percent of browser window or/and allow it to be resizable.

## To Reproduce
Just any bookmark.

## Screenshots
NA

## Notes
NA

Monirzadeh:
this pull request try to improve UX in wide screen or High-resolution display.
it should close #731 
for now it is not complete yet,
# TODO:
- [x] resize box
- [x] just active when display is bigger than full HD
- [x] remove `max-width` or change that to 800px if resize box is active

# Things we don't decide about them yet
- [ ] should new resize size remmeber?


fmartingr:
Check this:

```patch
diff --git internal/view/assets/less/custom-dialog.less internal/view/assets/less/custom-dialog.less
index f55a95c..243775a 100644
--- internal/view/assets/less/custom-dialog.less
+++ internal/view/assets/less/custom-dialog.less
@@ -19,18 +19,18 @@
        z-index         : 10001;
        background-color: rgba(0, 0, 0, 0.6);
        padding         : 32px;
-    resize          : both;
 
        .custom-dialog {
                display         : flex;
                flex-flow       : column nowrap;
                width           : 100%;
-               max-width       : 400px;
+               max-width       : 60vw;
                min-height      : 0;
-               max-height      : 100%;
+               max-height      : 70vh;
                overflow        : auto;
                background-color: var(--contentBg);
                font-size       : 16px;
+               resize          : both;
 
                .custom-dialog-header {
                        padding         : 16px;
@@ -54,6 +54,7 @@
                        align-content        : start;
                        align-items          : baseline;
                        grid-gap             : 16px;
+                       flex-grow            : 1;
 
                        &::after {
                                content          : "";
```

Also I got notice that we need to maintain both less and css, since there's no compiler in between. Maybe we can use less directly in the browser, do you want me to address that?

fmartingr:
> ```diff
> ```diff
> +                       flex-grow            : 1;
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> 
> i think i should fix this.
> 
> Update:
> maybe it is a little better. i don't like that empty space but not find a better solution yet.
> ```

Make it so the start size is the same as always, leaving the option to users to resize it to a larger viewport.

internal/view/assets/css/custom-dialog.css
fmartingr, line 1: Is this file necessary?

internal/view/assets/less/custom-dialog.less
```diff
@@ -153,4 +153,19 @@
 			}
 		}
 	}
-}
\ No newline at end of file
+}
+
+@media screen and (max-width: 400px) {
```
fmartingr, line 158: I've made a quick test and maybe we only need this `@media` query for mobile devices:

```less
@media only screen and (max-width: 768px) {
	.custom-dialog-overlay {
		padding: 0px;
	}

	.custom-dialog {
		width: 100%;
		height: 100%;
	}
}
```

This makes the dialog use the entire viewport, which seems useful mobile side:

Screenshot 2023-09-30 at 08 40 44

Don't get too distracted with mobile though, since there would be a lot of other things to improve in that side apart from this, just make it better without being to disruptive.

internal/view/assets/less/custom-dialog.less
```diff
@@ -153,4 +153,23 @@
...
+		padding: 0px;
+	}
+
+	.custom-dialog {
+		width: 100%;
+		height: 100%;
+	}
+}
+
+@media screen and (min-width: 769px) {
```
fmartingr, line 170: I think we don't need this one, it can be the dialog defaults, isnt' it?

internal/view/assets/less/custom-dialog.less
```diff
@@ -153,4 +153,23 @@
...
+
+
+@media only screen and (max-width: 768px) {
+	.custom-dialog-overlay {
+		padding: 0px;
+	}
+
+	.custom-dialog {
+		width: 100%;
+		height: 100%;
```
fmartingr, line 166: You are right that we can do this as well:

```suggestion
		height: 100%;
		resize: none;
```

internal/view/assets/less/custom-dialog.less
fmartingr: Can't we put those "defaults" directly in the `.custom-dialog` class and let the media query override them only for smaller screens? Maybe I'm wrong, haven't played with CSS in a while.

fmartingr (approved):

PR #736, opened 2023-09-16
deps: upgrade github.com/gofrs/uuid to v5

Monirzadeh:
`go-epub` will use uuid v5 #679  and we will have two version of dependency for same functionality. just update to v5


fmartingr (approved):
👍

PR #741, opened 2023-09-25
Switch to prefix search for bookmarks

### LINKED ISSUE #722: Search in readable content is unreliable
## Data
- 1.5.5
- SQLite(???). I did nothing to setup 
- Devuan x64 
- Web interface  

Search in readable content is unreliable.

## Expected behavior
Each & every work in readable content should be searchable.

## To Reproduce
1. Create "https://thehealthylife.com/" bookmark with archived content.
2. Look into readable content. See word "Markus" at the end of text.
3. Try searching for Markus
4. Observe "No saved bookmarks yet :("

## Screenshots
NA

## Notes
NA

rutkai:
Fixes #722.

It is not a 100% solution as I just added prefix matching, for pre- and postfix matching `LIKE` keyword would be needed, but that's a volume slower than `MATCH`.

### DIFF
--- a/internal/database/mysql.go
+++ b/internal/database/mysql.go
@@ -253,7 +253,7 @@ func (db *MySQLDatabase) GetBookmarks(ctx context.Context, opts GetBookmarksOpti
 			MATCH(title, excerpt, content) AGAINST (? IN BOOLEAN MODE)
 		)`
 
-		args = append(args, "%"+opts.Keyword+"%", opts.Keyword)
+		args = append(args, "%"+opts.Keyword+"%", opts.Keyword+"*")
 	}
 
 	// Add where clause for tags.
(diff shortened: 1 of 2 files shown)


rutkai:
It turned out tests are failing because I managed to find a bug in mysql/mariadb while trying to find a solution for the original issue: https://jira.mariadb.org/browse/MDEV-21750

As it's a bug, I'm not sure an input validation would help (good duct tape, though).

rutkai:
As an alternative fix for the original issue: prefix matching (and many other features of full-text search) can be displayed on the UI as tooltip or help so the user would include it in his/her search when prefix match is needed.

fmartingr:
Hey @rutkai, thank you for taking the time to look at this issue and dump your findings here, it would prove useful the moment much needed improvements are made into search :)

PR #742, opened 2023-09-29
refactor: migrate ebook routes

Monirzadeh:
this Pull request should remove legacy api for ebook.
# TODO :
- [x] API for download ebook
    - [x]  force user to login for use this API
    - [x] this will use for batch download ebook in UI
    - [x] other device can send request for specific bookmark and get epub file
    - [ ] don't get content from internet each time get that from database.
- [x] UI update to use new API
  - [x] if you select pdf article it take time. something slow down processes.
- [x] API for update ebook files
- [x] API for serve ebook files
- [x] Batch download use same route as update cache
- [x] remove part of code that no need them anymore
_this list will be update_ 
- [x] Update Swagger
- [ ] Add new unit test


fmartingr:
Hey @Monirzadeh, thinking about it I believe is best to move the download route outside of the API, since it does not return structured data (JSON), and that we use `GET` requests for the download so they are easier to call and can be easily cached in browsers as well.

We can have `/bookmarks/<id>/ebook/` to donwload a single ebook. I'm not sure about the bulk download, since I believe is not working right now, and we can use the frontend to call each download URL independently.

What do you think?

fmartingr:
So, if I get your comment. You want 

1) Migrate the API call to **generate** the ebooks of multiple bookmarks all at once. That would be `POST /api/v1/bookmarks/ebooks/ json={"bookmark_ids": ["id1", "id2"]}`. You sent an array of bookmarks IDs and bookmarks are generated on the server.
2) Ebooks are downloaded individually using `/bookmarks/<id>/ebook/`.

Is this correct?

I think my mistake was to think that the `/api/v1/bookmarks/ebooks/` was the endpoint that **downloaded** the ebooks.

fmartingr:
Love your diagram, it's super clean now :)

> should i use POST instead of PUT?

Ummm, I see it more as a `POST` since we are potentially creating content, but I'm 0/5 on that to be honest.

I will do a full review later in the day once I have some more focus time.

fmartingr:
Hey @Monirzadeh, I have been thinking a bit about this route.

Isn't this route the same as the update cache route (`/api/cache` currently) but only for ebooks? Could we migrate the cache route and use that to generate the ebooks sending the appropriate payload?

We could have:

```http
PUT /api/v1/bookmaks/cache
Content-Type: application/json
Authorization: Bearer xxx

{
  "bookmark_ids": [1,2,3],
  "generate_archive": bool,
  "generate_ebook": bool,
  "keep_metadata": bool,
}
```

Do we have this second endpoint for a specific reason?

fmartingr:
> ```httpspec
> /api/v1/bookmaks/cache
> ```

Mhmmm I don't agree with your first description.

For example if you want to download ebooks for bookmarks `[1,2,3]` as a client you would:

1. Request the ebooks to be generated: `PUT /api/v1/bookmaks/cache {"bookmark_ids": [1,2,3], "generate_ebook": true}`
2. Download each ebook: `GET /bookmark/<bookmark id>/ebook/` per bookmark.

We could even generate the ebook on the fly on the download ebook endpoint using a parameter, to generate it if is not generated previously: (`GET /bookmark/<bookmark id>/ebook/?generate=true` or something like that)

internal/http/routes/api/v1/bookmarks.go
```diff
@@ -20,11 +26,32 @@ type BookmarksAPIRoutes struct {
...
 	g.GET("/", r.listHandler)
+	g.PUT("/cache", r.updateCache)
 	g.POST("/", r.createHandler)
 	g.DELETE("/:id", r.deleteHandler)
 	return r
 }
 
+type updateCachePayload struct {
+	Ids           []int `json:"ids"    validate:"required"`
+	KeepMetadata  bool  `json:"keepMetadata"`
```
fmartingr, line 37: Now that we are creating new payloads, can we switch the JSON to `snake_case`?

internal/http/routes/api/v1/bookmarks.go
```diff
@@ -167,3 +194,147 @@ func NewBookmarksPIRoutes(logger *logrus.Logger, deps *config.Dependencies) *Boo
...
+		response.SendError(c, 404, "Bookmark not found")
+		return
+	}
+
+	if err != nil {
+		r.logger.WithError(err).Error("error getting bookmakrs")
+		response.SendInternalServerError(c)
+		return
+	}
+	// TODO: limit request to 20
```
fmartingr, line 248: **Note:** This can be done in the `updateCachePayload.IsValid()` method.

internal/http/routes/bookmark.go
```diff
@@ -66,3 +70,49 @@ func NewBookmarkRoutes(logger *logrus.Logger, deps *config.Dependencies) *Bookma
...
+	}
+
+	bookmark, found, err := r.deps.Database.GetBookmark(c, bookmarkID, "")
+	if err != nil || !found {
+		response.SendError(c, 404, nil)
+		return
+	}
+
+	if bookmark.Public != 1 && !ctx.UserIsLogged() {
+		response.SendError(c, http.StatusForbidden, nil)
```
fmartingr, line 107:
```suggestion
		response.SendError(c, http.StatusUnauthorized, nil)
```

internal/http/routes/bookmark.go
```diff
@@ -66,3 +70,49 @@ func NewBookmarkRoutes(logger *logrus.Logger, deps *config.Dependencies) *Bookma
...
+
+	bookmarkID, err := strconv.Atoi(bookmarkIDParam)
+	if err != nil {
+		r.logger.WithError(err).Error("error parsing bookmark ID parameter")
+		response.SendInternalServerError(c)
+		return
+	}
+
+	if bookmarkID == 0 {
+		response.SendError(c, 404, nil)
```
fmartingr, line 96:
```suggestion
		response.SendError(c, http.StatusNotFound, nil)
```

internal/http/routes/bookmark.go
```diff
@@ -66,3 +70,49 @@ func NewBookmarkRoutes(logger *logrus.Logger, deps *config.Dependencies) *Bookma
...
+	ctx := context.NewContextFromGin(c)
+
+	// Get server config
+	logger := logrus.New()
+	cfg := config.ParseServerConfiguration(ctx, logger)
+	DataDir := cfg.Storage.DataDir
+
+	bookmarkIDParam, present := c.Params.Get("id")
+	if !present {
+		response.SendError(c, 400, "Invalid bookmark ID")
```
fmartingr, line 84:
```suggestion
		response.SendError(c, http.StatusBadRequest, "Invalid bookmark ID")
```

internal/http/routes/bookmark.go
```diff
@@ -66,3 +70,49 @@ func NewBookmarkRoutes(logger *logrus.Logger, deps *config.Dependencies) *Bookma
...
+	}
+
+	if bookmarkID == 0 {
+		response.SendError(c, 404, nil)
+		return
+	}
+
+	bookmark, found, err := r.deps.Database.GetBookmark(c, bookmarkID, "")
+	if err != nil || !found {
+		response.SendError(c, 404, nil)
```
fmartingr, line 102:
```suggestion
		response.SendError(c, http.StatusNotFound, nil)
```

internal/http/routes/bookmark.go
```diff
@@ -66,3 +70,49 @@ func NewBookmarkRoutes(logger *logrus.Logger, deps *config.Dependencies) *Bookma
...
+	}
+
+	if bookmark.Public != 1 && !ctx.UserIsLogged() {
+		response.SendError(c, http.StatusForbidden, nil)
+		return
+	}
+
+	ebookPath := fp.Join(DataDir, "ebook", bookmarkIDParam+".epub")
+	if !ws.FileExists(ebookPath) {
+		response.SendError(c, 404, nil)
```
fmartingr, line 113:
```suggestion
		response.SendError(c, http.StatusNotFound, nil)
```

internal/view/login.html
```diff
@@ -121,6 +121,7 @@
 					}).then(json => {
 						// Save session id
 						document.cookie = `session-id=${json.message.session}; Path=${new URL(document.baseURI).pathname}; Expires=${json.message.expires}`;
+						document.cookie = `token=${json.message.token}; Path=${new URL(document.baseURI).pathname}; Expires=${json.message.expires}`;
```
fmartingr, line 124: Doesn't this make the previous line useless since you are overwritting it?

fmartingr (changes requested):
Hey, just reviewed this! Awesome work. I don't have much time these days to move things forward and all my time goes into reviews, so super happy seeing you contribute to the new API!

I left you some notes and changes around, and as a general comment I see you instantiate new loggers in handlers (or at least I saw one around), but know that there is one under `r.deps.Log` ;)

internal/view/login.html
fmartingr: > no it is not overwriting `session-id `just add `token` to the cookies. do we need keep `seasion-id` as cookies?

Yes, leave it there until we migrate all routes, just to be safe.

fmartingr (approved):
Awesome 👍

internal/http/routes/api/v1/bookmarks.go
fmartingr: Let's keep it like this for now, since the frontend can only show a limited amount of bookmarks anyway.

fmartingr:
> unit test is missing should i add that later in another pull request?

Yeah, is better that way so we don't keep coming back to the same changes 👍

PR #745, opened 2023-09-30
Adding a getting started doc page

jparrill:
Added a new Getting Started page with the basics of how to bootstrap the main functionality.
This allow the product to lower the entry level barrier


fmartingr:
Hey @jparrill, thanks for the contribution.

Have you checked the [usage#docker](https://github.com/go-shiori/shiori/blob/master/docs/Usage.md#running-docker-container) docs? Most of the content in this one is covered in there, maybe Getting Started is a better name for that part of the documentation?

Apart from that, I think you forgot to push the new screenshots.

jparrill:
Maybe it's better covered there, and a name change could help. From my PoV when I've checked the docs it was a bit difficult to just bootstrap the server with the parameters and then add the bookmarks. 

As you mention, changing the name of the Docker usage and putting (maybe a refernce) in the README.md could help on the visibility. Do you want me to follow this path or do you preffer to have a separated Getting Started section with the most basic usage of the app? w/e for me is fine :)

fmartingr:
Ummm, I would do a mix of what we have now. I believe that most of the content here can be divided into two sections:

- Usage: Would contain how to interact with Shiori in its two forms: web server and CLI.
- Getting started: is a way better name than Usage for newcomers, and most of the current content in Usage.md would apply here (installation). Then, depending on how the user wants to run it (CLI vs Server) we can link them to the proper usage anchor.

There is some work around thin over at #703, where I'm trying to restructure the documentation step by step.

What do you think?

docs/getting-started.md
```diff
@@ -0,0 +1,57 @@
...
+
+<!-- TOC -->
+
+- [Running Docker Container](#running-docker-container)
+- [How to use Shiori](#how-to-use-shiori)
+- [Deleting/Stopping Shiori](#deletingstopping-shiori)
+
+<!-- /TOC -->
+
+## Running Docker Container
```
fmartingr, line 13: I have been giving a thought and I believe is better to point the installation to the proper installation docs, so we don't have repeated documentation around. Hopefully we will maintain that properly and up to date and when the documentation site is up an ready it will be easy to navigate through sections. What do you think?

PR #747, opened 2023-10-03
fix: styles-check and swag-check monitor just needed directory not project root

### LINKED ISSUE #737: `make styles-check` return error if you have untrack files
## Data
- **Shiori version**: last master branch
- **Database Engine**: sqlite
- **Operating system**: linux
- **CLI/Web interface/Web Extension**: 

if you have files that not add to git to track  `make style-check` return error
```
fatal: No names found, cannot describe anything.
CLEANCSS_OPTS=--with-rebase /usr/bin/bash ./scripts/styles_check.sh
❌ git tree is not clean. Please commit all changes before running this script.
make: *** [Makefile:85: styles-check] Error 1
```

## Expected behavior
styles-check  return error if track files not commit.

## To Reproduce
1. Clone master project 
2. run `make style-check`
3. everything work fine
4. create an empty file in root of project test.sh
5. run `make style-check` again
6. you see that error

## Notes
we have some files in project that not ready to add yet or some other is personal script that we use for easier development processes things that we not add to project or `.gitignore` so in this situation we can't use `make style-check` command.

Monirzadeh:
currently if we use `make styles-check` or `make swag-check` those scripts return error if we have uncommit file even if they are not related to CSS or swagger.
this try to fix #737

### DIFF
--- a/scripts/styles_check.sh
+++ b/scripts/styles_check.sh
@@ -3,15 +3,15 @@
 # This script is used to check if the style.css file is up to date.
 
 # Check if the git tree for CWD is clean
-if [ -n "$(git status internal/view/assets/less internal/view/assets/css --porcelain)" ]; then
+if [ -n "$(git status internal/view/assets/css --porcelain)" ]; then
     echo "❌ git tree is not clean. Please commit all changes before running this script."
     exit 1
 fi
 
 # Check style.css file
 CLEANCSS_OPTS=${CLEANCSS_OPTS} make styles
-if [ -n "$(git status internal/view/assets/less internal/view/assets/css --porcelain)" ]; then
+if [ -n "$(git status internal/view/assets/css --porcelain)" ]; then
     echo "❌ style.css wasn't built from less changes. Please run 'make styles' and commit the changes."
-    git reset --hard
+    git checkout -- internal/view/assets/css/
     exit 1
 fi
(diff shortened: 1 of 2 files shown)


scripts/styles_check.sh
```diff
@@ -3,14 +3,14 @@
 # This script is used to check if the style.css file is up to date.
 
 # Check if the git tree for CWD is clean
-if [ -n "$(git status --porcelain)" ]; then
+if [ -n "$(git status internal/view/assets/less internal/view/assets/css --porcelain)" ]; then
```
fmartingr, line 6: I believe we only need to check the css path here and below, since that's the generated code.

scripts/styles_check.sh
```diff
@@ -3,14 +3,14 @@
...
     echo "❌ git tree is not clean. Please commit all changes before running this script."
     exit 1
 fi
 
 # Check style.css file
 CLEANCSS_OPTS=${CLEANCSS_OPTS} make styles
-if [ -n "$(git status --porcelain)" ]; then
+if [ -n "$(git status internal/view/assets/less internal/view/assets/css --porcelain)" ]; then
     echo "❌ style.css wasn't built from less changes. Please run 'make styles' and commit the changes."
     git reset --hard
```
fmartingr, line 15: This would reset all repository to the previous state, instead of only the generated files, I think using `git checkout` here is better now.

fmartingr (changes requested):
Hey, quickly reviewed and made some comments.

Also the swag part seems missing?

fmartingr (approved):
Awesome, thanks for the fix!

PR #748, opened 2023-10-10
feat: allow authentication using cookies

fmartingr:
Allow the `AuthenticationMiddleware` to read the JWT token from cookies so we can load static pages outside the API (like the bookmark archive and ebook) safely.


fmartingr:
Included in https://github.com/go-shiori/shiori/pull/742

PR #759, opened 2023-10-15
fix: Actions in overlays on mobile hard to press

cbe:
Buttons on mobile devices could've been cut off, depending on the browsers chrome size which rendered them either completely off-screen or at least hard to reach. Using `dvh` (dynamic viewport height) should fix this issue for most newer browsers.

:information_source: I could not find an issue for this problem, do you want me to create one so it's properly tracked?

Also: these types of units are rather new (stable since mid to end of last year, as seen on [caniuse.com](https://caniuse.com/viewport-unit-variants)). What's your opinion on fallback values to the original `vh` for example?

## Screenshots

|`master`|with fix|
| --- | --- |
|before|with-fix|


fmartingr:
Hey @cbe, nice find!

> ℹ️ I could not find an issue for this problem, do you want me to create one so it's properly tracked?

There's no need, the PR is enough.

> Also: these types of units are rather new (stable since mid to end of last year, as seen on [caniuse.com](https://caniuse.com/viewport-unit-variants)). What's your opinion on fallback values to the original `vh` for example?

If that's not very cumbersome, it would be nice to have, specially for people in mobile which are the target of this PR.

Thanks for the contribution and taking the time to fix this!

cbe:
> If that's not very cumbersome, it would be nice to have, specially for people in mobile which are the target of this PR.

@fmartingr not cumbersome at all, I added some fallback values so it works for older browsers as well. I just pushed the commit providing fallback values.

> Thanks for the contribution and taking the time to fix this!

You're welcome :slightly_smiling_face: I like Shiori

fmartingr (approved):
Awesome fix, thank you! The other day I was adding a bookmark with my phone and I used [Enter] instinctively, but noticed that the add button wasn't there. Hopefully this will fix things :)

cbe:
> Awesome fix, thank you! The other day I was adding a bookmark with my phone and I used [Enter] instinctively, but noticed that the add button wasn't there. Hopefully this will fix things :)

Thanks for merging this :bow:! I also got another fix up and ready for the case you mentioned, funny coincidence!

PR #760, opened 2023-10-15
#WIP: Webroot support on new server

fmartingr:
**Note**: I'm not sure if this will reach production, since using a reverse proxy is an user's control and most reverse proxies should have options to send the unprefixed path to their upstreams, rather than have all services underneath handle a prefix.


fmartingr:
Closing in favor of instructing users in how to properly setup their reverse proxies.

Issue #369, opened 2022-02-13
Unify webextension and regular APIs


fmartingr:
After making some changes I noticed that there are two APIs, one for the frontend/users and one for the web extension. I need to dig further into why is this way, but unifying both APIs seems the way to go.

Relates to: #184

PR #761, opened 2023-10-15
add create ebook by default in settings

### LINKED ISSUE #720: Add user setting to determine if an ebook should be created by default
as part of #682 
we need an option to generate ebook by default it should work in webui
it is will waiting for #639 @fmartingr please set this somewhere proper after v1.6.1

Monirzadeh:
add a settings for active generate ebook by default.
# TODO:
- [x] add that to settings and work in update 
- [x] create ebook work on add new bookmark
- [x] ebook should generate after constant is available ~~(new bookmark create empty ebook right now)~~

## What is not done in this PR
- web extension and command line until we finish work on #369 

fix #720

### DIFF
--- a/internal/core/processing.go
+++ b/internal/core/processing.go
@@ -142,6 +142,7 @@ func ProcessBookmark(req ProcessRequest) (book model.Bookmark, isFatalErr bool, 
 	// If needed, create ebook as well
 	if book.CreateEbook {
 		ebookPath := fp.Join(req.DataDir, "ebook", strID+".epub")
+		req.Bookmark = book
 
 		if strings.Contains(contentType, "application/pdf") {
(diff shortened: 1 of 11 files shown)


internal/core/processing.go
```diff
@@ -142,6 +142,7 @@ func ProcessBookmark(req ProcessRequest) (book model.Bookmark, isFatalErr bool,
 	// If needed, create ebook as well
 	if book.CreateEbook {
 		ebookPath := fp.Join(req.DataDir, "ebook", strID+".epub")
+		req.Bookmark = book
```
fmartingr, line 145: What does this solve?

fmartingr (approved):

PR #765, opened 2023-10-22
Make suggestions tapable/clickable

cbe:
This was especially noticeable when using the web app on mobile devices as the suggestion looks like it's tapable. It should be also a little bit more convenient UX-wise when it's allowed to be clicked in non- mobile context as well.


fmartingr (approved):
Tested this on my phone, and while the suggestion list shows up in the middle of the screen, tapping works and it's more convenient that are the bottom, where it may add more scrolling.

PR #794, opened 2023-12-02
Add Documentation>CLI>Add bookmark

LLKoder:
Added `Add bookmark` instruction and examples in CLI documentation.


fmartingr (approved):
Thanks for the contribution @LLKoder! Hopefully we can automate this in the future so we don't need to manually update the documentation when we change the CLI.

PR #802, opened 2023-12-22
fix: use noreferer to prevent exposing shiori instance url to archived websites

istiak101:
https://caniuse.com/rel-noreferrer

### DIFF
--- a/internal/view/assets/js/component/bookmark.js
+++ b/internal/view/assets/js/component/bookmark.js
@@ -4,7 +4,7 @@ var template = `
 		v-if="editMode" 
 		@click="selectBookmark">
 	</a>
-	<a class="bookmark-link" :href="mainURL" target="_blank" rel="noopener">
+	<a class="bookmark-link" :href="mainURL" target="_blank" rel="noopener noreferrer">
 		<span class="thumbnail" v-if="thumbnailVisible" :style="thumbnailStyleURL"></span>
 		<p class="title" dir="auto">{{title}}
 			<i v-if="hasContent" class="fas fa-file-alt"></i>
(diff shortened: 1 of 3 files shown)


Monirzadeh:
should we do that in this line too?
https://github.com/go-shiori/shiori/blob/188ce68761eada733d2fc905df71ca3f68de4058/internal/view/content.html#L29

istiak101:
> should we do that in this line too?

I haven't found its usage yet in the application. If it goes to a url in browser then yes. Can you please confirm?

Monirzadeh:
as i see you add `noreferrer` for link to the bookmark link in home page. right?
we have other link too. for example in content page:
https://github.com/go-shiori/shiori/blob/188ce68761eada733d2fc905df71ca3f68de4058/internal/view/content.html#L29-L36
or
https://github.com/go-shiori/shiori/blob/188ce68761eada733d2fc905df71ca3f68de4058/internal/webserver/handler.go#L74-L84

why you not add `noreferrer` to them?

istiak101:
I downloaded the stable release which is 49 commits behind. I will build main and check again.

fmartingr:
As @Monirzadeh pointed out, there are other places where we link to the original bookmark URL, but in general I think this is a good addition. Ping me if you need help pinpointing where all the links are in the code and thank you for the contribution!

istiak101:
Thanks @Monirzadeh for pointing out missing noreferrer value.

@fmartingr I have updated the PR. I think these are only places where noreferrer is required.

fmartingr (approved):
I think you covered all links with this, thanks for your contribution!

fmartingr:
> LGTM 👍 Thanks for PR we can open an issue for
> 
> > Edit: bookmark/ID/content and bookmark/ID/archive/ would still leak shiori servers url. I am not familiar with the project but I think this is done with https://github.com/go-shiori/dom .

I think he solved most of the links around. Since we need to work a bit on archives in the future I'd say this is a non-issue for us until we get a contributor to handle this.

PR #836, opened 2024-02-05
feat: allow authentication using proxy request header

PterX:
This submission implements the working method of oauth header authentication, which needs to be combined with the reverse proxy service and oauth service to achieve password-free authentication login.
For example: The following is explained by combining the configuration of nginx and authelia
Refer to the auth documentation page https://www.authelia.com/integration/proxies/nginx/

We can use the following configurations
/config/nginx/snippets/proxy.conf
/config/nginx/snippets/authelia-location.conf
/config/nginx/snippets/authelia-authrequest.conf

nginx configuration is as follows
```
server{
    listen 80;
    server_name shiori.example.com;

    include /config/nginx/snippets/authelia-location.conf;

    location / {
        proxy_pass http://10.0.0.1:8080; // pass to shiori
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection 'upgrade';
        include /config/nginx/snippets/authelia-authrequest.conf;
        include /config/nginx/snippets/proxy.conf;
    }
}
```

Way of working
step 1
Client  request -> nginx proxy(auth request, not login) -> authelia ( redirect to auth page)  

step 2
Client  request -> nginx proxy(auth request, it was login with oauth) -> authelia (verify api) -> shiori (get auth user header, like user name) -> shiori (getTokenFromAuthHeader:  trust && query account by user name) -> shiori (generate token and session-id,save to cookie)

step 2
Client request(with token,session-id) -> nginx proxy(also auth request,ignore) -> shiori (old work path, getTokenFromCookie, and other check ...) -> shiori (user verify ok) -> shiori ( business code)

Closes #594

PR #837, opened 2024-02-06
fix: regressions and documentation from 1.5.5 upgrade

fmartingr:
- Add required flags for the `serve` command to behave properly after upgrading to `1.6.0`.
- Updated initial account references in the docs, since now it's a real account it wont be removed automatically.

### DIFF
--- a/docs/Frequently-Asked-Question.md
+++ b/docs/Frequently-Asked-Question.md
@@ -2,21 +2,17 @@
 
 <!-- TOC -->
 
-- [Frequently asked questions](#frequently-asked-questions)
-- [General](#general)
-    - [What is this project ?](#what-is-this-project-)
-    - [How does it compare to other bookmarks manager ?](#how-does-it-compare-to-other-bookmarks-manager-)
-    - [What are the system requirements ?](#what-are-the-system-requirements-)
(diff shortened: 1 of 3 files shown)


fmartingr:
Hey @Monirzadeh take a look at this, anything I missed and should take a look at before releasing 1.6.0? I did a 1.5.5 to 1.6.0 upgrade locally and everything seem working as usual.

fmartingr:
@Monirzadeh Thanks for the suggestions. Answering to your comments:

- The `shiori serve` still works, but shows as deprecated when run and is only a proxy of the `shiori server` command. We will remove it at a later date.
- I will update all references to `shiori serve` from the docs. Good catch.
- I will remove the commented out old api endpoints
- I will update the go-epub dependency (and any other outdated dependencies)
- #836 won't be ready for 1.6.0 and I don't want to put anything else for the release, it has enough content as is.

fmartingr:
> Beside you forget to remove `# General` line (That don't have that in toc) things ok to me 👍

Oh sorry, I forgot about that. Fixed the headings https://github.com/go-shiori/shiori/pull/837/commits/d679c844991595ca440d7acb178941e1a9d00b54

PR #858, opened 2024-03-15
fix: thumbnail should show last thumbnail in server in any condition

### LINKED ISSUE #857: Issue with image display in Chrome v1.6.0-rc.7
### Problem Description
I am experiencing an issue with the display of images in the Shiori app when using Google Chrome. I'm currently testing with version v1.6.0-rc.7 of Shiori.

### Steps to Reproduce
1. Create several bookmarks in the app.
2. Delete all the created bookmarks.
3. Add several bookmarks again.
4. Open the app in an incognito tab on the left and a normal tab on the right (See capture) 
5. Observe that on the right tab, only one image is displayed correctly.

Captura de pantalla_área-de-selección_20240312164127

Edit: I've also noticed that this issue occasionally occurs in incognito mode as well. 

image

Monirzadeh:
thumbnail not update in browser sometimes.

`bun.lockb` update with `make style` so i just push that too.

Closes #857
Depends on #896


internal/view/assets/js/component/bookmark.js
```diff
@@ -94,8 +94,9 @@ export default {
 			return this.excerpt !== "" && !this.thumbnailVisible && !this.HideExcerpt;
 		},
 		thumbnailStyleURL() {
+			var cacheBuster = new Date().getTime();
 			return {
-				backgroundImage: `url("${this.imageURL}")`,
```
fmartingr, line 98: Can we use some properties from the bookmark instead of a random string? This could break the cache unintentially since each time the query string is going to be different. Can we use `?mtime=<bookmark modified time>` ?

fmartingr (changes requested):

internal/view/assets/js/component/bookmark.js
fmartingr: I always assumed `Bookmark.ModifiedTime` was the datetime when the bookmark was last modified (attributes/archive/readable/ebook) modified, it seems that is not the case. Let me think and get back to you.

fmartingr: I ended up not updating you here.

My idea is:
- Refactor the migration system (#876)
- Add a `created` column to the bookmarks with the time the bookmark was first created and prefill it with the current `modified` for existing bookmarks.
- Update `modified` every time the bookmark is updated (either the data, content, archive, etc).

Then, we can use the `modified` as query parameter for the thumbnail.

fmartingr:
I'm thinking that instead of the query parameter, what do you think if we send the `Last-Modified` and `ETag` headers in the response?

```
Last-Modified: {formatted modified_at}
ETag: W/{filepath}-{modified_at}
```

What do you think?

fmartingr:
Can you refactor `SendFile` adding a nilable argument with options coming from a struct?


Something along the lines:
```
type SendFileOptions struct {
  Headers []http.Header
}


func SendFile(c *gin.Context, storageDomain model.StorageDomain, path string, options *SendFileOptions) {
...
}
```

And if we don't provide `Cache-Control`/`ETag` headers in the options, defaults are set in the function.

What do you think?

fmartingr:
If the best course of action here is to provide a queryparameter with the modified time, let's go for that so we fix the bug. Any improvements can be handled later.

fmartingr (approved):
👍

PR #859, opened 2024-03-17
Add securityContext to provided Kubernetes manifest

### LINKED ISSUE #791: security context is necessary for Kubernetes
This is really just a documentation issue. If you're deploying to k8s and not using hostPath but a real persistent storage (i.e. Longhorn) you'll need to set the security context to match the user you created in the Dockerfile. Otherwise you get a permissions issue and the `Failed to open database: unable to open database file: out of memory (14)` error from sqlite 

``` 
spec:
      securityContext:
        runAsUser: 1000
        runAsGroup: 1000
        fsGroup: 1000
      volumes:
```

piano1029:
In the (currently) provided Kubernetes deployment manifest the securityContext is missing, this results in Shiori crashing with a vague "out of memory" error. This pull request adds the securityContext with the same uid and gid values as in the Docker image (1000), fixes #791.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -83,6 +83,10 @@ spec:
       labels:
         app: shiori
     spec:
+      securityContext:
+        runAsUser: 1000
+        runAsGroup: 1000
+        fsGroup: 1000
       volumes:
       - name: app
         hostPath:


fmartingr:
Hey @piano1029, thanks for the contribution. I will probably update the deployment docs with a link to an helm chart instead of having that YAML around, the docs require some love in all areas. Thanks for catching this.

PR #865, opened 2024-03-23
fix: override configuration from flags only if set

fmartingr:
This pull request checks if the flags provided to the `server` command has been set (are changed from the default values) before overriding the configuration with their values.

This fixes the configuration not being able to work when using only environment variables, since the default values were always overriden by the flags, regardless of being set up or not.

Ping @tenpai-git

Closes #844

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -4,14 +4,13 @@ about: Report problems with Shiori
 title: One line description of the bug
 labels: type:bug
 assignees: ''
-
 ---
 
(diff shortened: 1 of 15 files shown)


tenpai-git (approved):
Thank you @fmartingr! This looks good to me.

tenpai-git (approved):

tenpai-git (approved):

fmartingr:
> LGTM and work fine in my system can you check codecov error before merge this it decrease coverage

Going to merge this for now so we can release a new dot version. I need to take a look at viper or something that allows us to have both flag and envvars easily.

PR #868, opened 2024-03-24
fix: incorrect endpoint in rename tags operation

### LINKED ISSUE #867: Cannot rename tags in v1.6.0
## Data
- **Shiori version**: 1.6.0
- **Database Engine**: SQLite
- **Operating system**: Debian (Docker)
- **CLI/Web interface/Web Extension**: ?

No matter which tag I rename, the backend always returns an HTTP 404 error.

## Expected behavior
The new name should be applied and the HTTP response code should be 200.

## To Reproduce
1. Click on the button titled: "Show tags"
2. Click on "Rename Tags"
3. Select any tag
4. Change its name and click on "OK"
5. See error

## Screenshots
image
image

Zlendy:
fixes #867

### DIFF
--- a/internal/view/assets/js/page/home.js
+++ b/internal/view/assets/js/page/home.js
@@ -953,7 +953,7 @@ export default {
 					};
 
 					this.dialog.loading = true;
-					fetch(new URL("api/tag", document.baseURI), {
+					fetch(new URL("api/tags", document.baseURI), {
 						method: "PUT",
 						body: JSON.stringify(newData),
 						headers: {


fmartingr:
Thank you for noticing this @Zlendy! It seems I made a typo when migrating the legacy routes from the old handlers, [the old API used a mix of `/api/tag` and `/api/tags`...](https://github.com/go-shiori/shiori/blob/master/internal/http/routes/legacy.go#L85-L86). Anyway, let me merge this so we can cut a patch release this week with some fixes :)

fmartingr (approved):

PR #872, opened 2024-03-31
fix: api calls using absolute urls

### LINKED ISSUE #870: Some API calls do not respect root path
## Data
- **Shiori version**: 1.6.0
- **Database Engine**: MySQL
- **CLI/Web interface/Web Extension**: Web interface

Some API calls in the settings section of the front end don't use the base path. This includes:

- [Saving any setting](https://github.com/go-shiori/shiori/blob/5a166906db813917b4db23f410d929d499ad5365/internal/view/assets/js/page/setting.js#L110)
- [Deleting an account](https://github.com/go-shiori/shiori/blob/5a166906db813917b4db23f410d929d499ad5365/internal/view/assets/js/page/setting.js#L340)

## Expected behavior
I expect my settings to save!

## To Reproduce
1. Set up an installation on a root path, ex: `/shiori`. 
2. Log in an admin account and navigate to settings.
3. Click on 'Show bookmark's ID'

## Screenshots
The error message is the HTML of another service I run, so I'd prefer to keep that private.

## Notes
I dug around in the code a bit, and the linked lines I believe are the culprit. It looks like there's some inconsistency with API fetches, specifically the initial `/` for settings and lack of `URL` entirely for deleting accounts.

I can whip up a PR if needed.

fmartingr:
Some API calls were using an absolute path ignoring the `base href` element thus not working/pointing where it shouldn't.

Fixes #870

### DIFF
--- a/internal/view/assets/js/page/setting.js
+++ b/internal/view/assets/js/page/setting.js
@@ -107,7 +107,7 @@ export default {
 
 			this.$emit("setting-changed", options);
 			//request
-			fetch(new URL("/api/v1/auth/account", document.baseURI), {
+			fetch(new URL("api/v1/auth/account", document.baseURI), {
 				method: "PATCH",
 				body: JSON.stringify({
 					config: this.appOptions,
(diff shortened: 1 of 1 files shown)


fmartingr:
> 👍 why this should not happen for swagger? endpoint in there start with `/` too . dose swagger handle that automatically? can confirm swagger work fine in this situation too?

I'm not sure tbh. But let me merge this one and address swagger in a separate one if reported.

PR #878, opened 2024-04-07
fix: replace --rm-dist with --clean

Monirzadeh:
gorelease update because  `--rm-dist` deprecated
https://goreleaser.com/deprecations/#-rm-dist


fmartingr (approved):
👍

PR #881, opened 2024-04-08
Add PikaPods hosting option

m3nu:
Add *Managed Hosting* section and PikaPods as provider.


fmartingr (approved):
Thank you @m3nu, will be awesome to see more people use your services.

Issue #851, opened 2024-02-29
Windows Binaries are being released as .tar.gz's


torculus:
Just like it says, the releases page distributes Windows `.exe`'s in `.tar.gz` files. These can't be opened natively on Windows, and it's blocking submission of `shiori` to `winget`

Here's the relevant section in goreleaser:
https://github.com/go-shiori/shiori/blob/595cb45a2ce7031926ead1f1e1324e673e36d676/.goreleaser.yaml#L29

Another golang project I follow [solves this](https://github.com/twpayne/chezmoi/blob/master/.goreleaser.yaml#L99) with this setting
```yaml
  format_overrides:
  - goos: windows
    format: zip
```

torculus:
@Monirzadeh no, for several reasons:
1. Most importantly: effort. Releasing as a `.zip` requires minimal effort, while releasing for `winget` requires registering with the Microsoft contributor covenant and (possibly) creating a new git branch or repo just for `winget` manifest updates
2. The `winget` manifest schema changes every few months, and there are plenty of broken/barely-working packages in `winget` which are the result of automated updates
3. There may be some Windows users who can't/don't want to use winget. Distributing as a `.zip` also allows for people to test `shiori` without installing it

Although it is possible to overcome each of those issues, releasing as a `.zip` is probably a good place to start ***right now***. Winget support can always be added later 🙂

fmartingr:
Closing this as it was fixed in #884

PR #884, opened 2024-04-08
Update .goreleaser.yaml

torculus:
This addresses #851. I tested with `goreleaser` and it created the Windows `.zip` archive, but couldn't publish because I had invalid `GITHUB_TOKEN` credentials


fmartingr (approved):
Awesome, thank you for this and glad to see it was that easy. Sorry that it fell out of my radar.

PR #885, opened 2024-04-08
feat: endpoint to serve readable content

Monirzadeh:
this Pull request serve readable content (text only and html) on `/api/v1/bookmarks/id/readable`
- [x] serve HTML and text of bookmarks
- [x] be sure it authenticate correctly
- [x] fix bug related to send duplicate response if user not authenticate
   *bug found wait for fix*
- [x] add unit test


fmartingr:
A fix for the auth middleware: https://github.com/go-shiori/shiori/pull/888

internal/http/routes/api/v1/bookmarks.go
```diff
@@ -57,6 +58,63 @@ func (p *updateCachePayload) IsValid() error {
...
+	}
+
+	bookmark, err := r.deps.Domains.Bookmarks.GetBookmark(c.Context, model.DBID(bookmarkID))
+	if err != nil {
+		response.SendError(c.Context, http.StatusNotFound, nil)
+		return nil, model.ErrBookmarkNotFound
+	}
+
+	return bookmark, nil
+}
```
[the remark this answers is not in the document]
fmartingr: What code block exactly, this entire function?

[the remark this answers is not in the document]
fmartingr: We can think about it later when the entire API is in place, don't worry too much now.

internal/http/routes/api/v1/bookmarks.go
```diff
@@ -57,6 +58,63 @@ func (p *updateCachePayload) IsValid() error {
 	return nil
 }
 
+type contentResponseMessage struct {
```
fmartingr, line 61:
```suggestion
type readableResponseMessage struct {
```

internal/http/routes/api/v1/bookmarks.go
```diff
@@ -57,6 +58,63 @@ func (p *updateCachePayload) IsValid() error {
...
+//	@Failure					403	{object}	nil	"Token not provided/invalid"
+//	@Router						/api/v1/bookmarks/id/readable [get]
+func (r *BookmarksAPIRoutes) bookmarkReadable(c *gin.Context) {
+	ctx := context.NewContextFromGin(c)
+
+	bookmark, err := r.getBookmark(ctx)
+	if err != nil {
+		return
+	}
+	responseMessage := contentResponseMessage{
```
fmartingr, line 110:
```suggestion
	responseMessage := readableResponseMessage{
```

internal/http/routes/api/v1/bookmarks.go
fmartingr: Can you move this struct closer to the handler that uses it? (right on top of it)

fmartingr (changes requested):
I like how you called the endpoint "readable" instead of "content", we should probably do that in the code as well. I always confused "content" with "archive". Readable is a better name for this.

fmartingr (approved):

PR #886, opened 2024-04-10
Update docker-compose.yaml

MikePadge:
Howdy, I commented out the rest of the mariadb declaration, and added a comment, since the container does not need to start if postgres is the active default DB for the example. 

I also added a volume declaration for postgres to add some default persistence. Might save some new users from running a compose down and wiping their data.  

These are just opinions, feel free to accept or not. ¯\_(ツ)_/¯ 
Thanks for this project, it's pretty awesome.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -13,8 +13,9 @@ services:
       - ".:/src/shiori"
     restart: unless-stopped
     links:
+    # Comment/Uncomment the postgres mariadb sections depending on the database you choose.
       - "postgres"
-      - "mariadb"
+      #- "mariadb"
     environment:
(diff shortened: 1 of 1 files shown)


fmartingr:
Hey @MikePadge, thanks for the contribution. Just for a bit of information, this docker-compose is not meant to use in production environment and that's why it has the both services enabled, we use that locally to run shiori using other database engines apart from SQLite and to run the test suite, that's why it didn't have volumes and the shiori service is running from a dockerfile rather than use an image in the registry.

So my suggestion for this would be lo leave both services as it where and comment out the volume parts of psql (and add one to mysql if you can) with a note for people to uncomment it if they need persistence. What do you think?

PR #888, opened 2024-04-13
fix: authentication flow should abort early

fmartingr:
- Ensures that the connection is aborted from the Auth middleware when the authentication is unsuccessful.
- Modified the main middleware test to ensure that we are not filtering out extra response data
- Added helper methods to `testutil.TestResponse`
- Fixed some typos
- Fixed tests for the tags and bookmarks api v1

### DIFF
--- a/internal/http/middleware/auth_test.go
+++ b/internal/http/middleware/auth_test.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/gin-gonic/gin"
+	"github.com/go-shiori/shiori/internal/http/response"
 	"github.com/go-shiori/shiori/internal/model"
 	"github.com/go-shiori/shiori/internal/testutil"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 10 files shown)


fmartingr:
> Hi if you send an unauthorized request it will not reach to the any endpoint in `internal/http/routes/api/v1/bookmarks.go` Why is it important? in #885 for public bookmarks you can't accesses readable content bemuse your request not reach to the endpoint (if be unauthorized) you can test that in other way send a update cache request to the
> 
> ```
> http://127.0.0.1:8080/api/v1/bookmarks/cache
> ```
> 
> if you are unauthorized than you get 401 but this status code not came from this line
> 
> https://github.com/go-shiori/shiori/blob/c77a542a5b450b496d0def40acacd31874a647da/internal/http/routes/api/v1/bookmarks.go#L70-L73
> 
> it force us to all api in `internal/http/routes/api/v1/bookmarks.go` be authorize (and public status be pointless on endpoint in there) dose you want this?

Right now all handlers under the bookmark endpoint require authentication, so that's good as it is. When we add new mixed handlers we could add the middleware just above the endpoints that require it. Is this something you need for your PR?

fmartingr:
@Monirzadeh Check https://github.com/go-shiori/shiori/pull/888/commits/dff17e1529426f63b03c2cf5c0b18c5c86aa4db9

fmartingr:
> @Monirzadeh Check [dff17e1](https://github.com/go-shiori/shiori/commit/dff17e1529426f63b03c2cf5c0b18c5c86aa4db9)

Ugh, now I remember why I did it the other way, routes were easier to test.

PR #894, opened 2024-04-27
fix(sqlite): id column should be unique

### LINKED ISSUE #892: Shiori auto increasment not work currectly on sqlite database
## Data
- **Shiori version**: 1.6.2
- **Database Engine**: sqlite
- **Operating system**: linux
- **CLI/Web interface/Web Extension**: all of them

if you add just first bookmark it will get ID 1
now if you remove that bookmark and add new bookmark it again get same ID number 1

## Expected behavior
shiori always should create uniq id that not ever use.

## To Reproduce
1. add bookmark a that you will get id 1 for that
2. delete that bookmark
3. add bookmark b
4. you will get id 1 again instead of 2 for bookmark b

## Notes
it happen just in sqlite

Monirzadeh:
This PR try to fix #892


fmartingr (approved):
LGTM. Did you test the migration with data manually on your end?

fmartingr:
> today i thinking about a specific situation. should we handle id that currently reuse in database?

I don't think so. It already happened and we are avoiding it in the future now.

PR #896, opened 2024-05-01
feat: store created and modified time separately on database for bookmarks

Monirzadeh:
this PR should done two things:
1. add `created` time to the database
2. `modified` update if write new content in an exist bookmark.
*any place you will done change in any aspect of bookmark (new archive,content,epub,tags,etc)*

in future if we want to update modified time any place in code just set `book.modified = ""` shiori automatically will update modified time.

# TODO:
- [x] sqlite migrate
- [x] update model
- [x] fix logic of update modified time
- [x] show added time in content page correctly
- [x] repeat that for postgres database 
- [x] change mariadb update `modfied_at` same as other database *mariadb update modified time when row change that not possible in other database* this change in this PR and all use same method now.
- [x] repeat that for mariadb database 
- [x] add unit tests


fmartingr:
Hey @Monirzadeh is this ready for review? I see items pending.

A side comment for me, would it me much effort renaming the `modified` column to `modified_at` and use `created_at` as creation date column?

fmartingr:
> mysql is not ready yet when i fill created with modified value (in maraidb) created (around one or two) second different with modified and i can't find why it happen yet.

Let me know if you get stuck with this, I can take a look if you want.

> ok i will change names column to `modified_at` and `created_at` should i change that in API too or keep that untouch?

Leave the old API as is, we will use the new field in the new API.

internal/view/content.html
```diff
@@ -38,6 +37,9 @@
 		<div id="content" dir="auto" v-pre>
 			$$.HTML$$
 		</div>
+        <div id="footer">
+            <p id=metadata> {{ createdtime() === modifidetime() ? `Added ` +  createdtime() : `Added ` + createdtime() + '  |  Last Modified ' + modifidetime() }} </p>
```
fmartingr, line 41: Shouldn't something like this be more clear?

```suggestion
            <p id=metadata>Added {{ createdtime() }} {{ createdtime() !== modifidetime() ? `Added ` + createdtime() + '  |  Last Modified ' + modifidetime() }} </p>
```

internal/view/content.html
```diff
@@ -48,15 +50,26 @@
...
 			data: {
-				modified: "$$.Book.Modified$$"
+				created: "$$.Book.Created$$"
 			},
 			methods: {
-				localtime() {
-					var strTime = this.modified.replace(" ", "T");
-					if (!strTime.endsWith("Z")) {
-						strTime += "Z";
+				createdtime() {
```
fmartingr, line 56: Can we use `camelCase` here?
```suggestion
				createdTime() {
```

internal/view/content.html
```diff
@@ -48,15 +50,26 @@
...
+					if (!strCreatedTime.endsWith("Z")) {
+						strCreatedTime += "Z";
 					}
-					return dayjs(strTime).format("D MMMM YYYY, HH:mm:ss");
+					if (!strModifiedTime.endsWith("Z")) {
+						strModifiedTime += "Z";
+					}
+                    return `${dayjs(strCreatedTime).format("D MMMM YYYY")}`;
+				},
+				modifidetime() {
```
fmartingr, line 67: Can we use `camelCase` here? (also there's a typo)

```suggestion
				modifiedTime() {
```

internal/webserver/handler-api.go
```diff
@@ -360,6 +360,9 @@ func (h *Handler) ApiUpdateBookmark(w http.ResponseWriter, r *http.Request, ps h
 		}
 	}
 
+	// Set bookmark modifide
```
fmartingr, line 363:
```suggestion
	// Set bookmark modified
```

internal/webserver/handler-api.go
```diff
@@ -360,6 +360,9 @@ func (h *Handler) ApiUpdateBookmark(w http.ResponseWriter, r *http.Request, ps h
...
+	book.Modified = ""
```
fmartingr, line 364: Why empty?

fmartingr (changes requested):
Hey @Monirzadeh, good job as ususal. I've left a few comments around for changes.

Also, I'm missing tests for:

- Ensuring created_at time is set correctly when creating
- Ensuring modified_at time is set correctly when updating
- Ensuring created_at filter works properly
- Ensuring modified_at filter works properly

Let me know if you need help with any of those.

Thank you!

fmartingr:
Forgot to check, but we should probably add indexes in the datetime fields since we are going to rely on those heavily on results.

fmartingr:
> > Forgot to check, but we should probably add indexes in the datetime fields since we are going to rely on those heavily on results.
> 
> something like this?
> 
> ```
> CREATE INDEX idx_created_at ON bookmark(created_at);
> CREATE INDEX idx_modified_at ON bookmark(modified_at);
> ```

Yes, ensuring the fields have indexes so queries work performant in databases, though it will take more disk space.

> can you please give me an example for this
> 
> > Ensuring modified_at filter works properly

We are adding the created_at/modified_at fields because we want to be able to query the API using, so we need to check that we can call `GetBookmarks` with proper options to get the items sorted by both fields. I believe we have `ByLastAdded` and `ByLastModified` now, but `ByLastAdded` uses the `id` instead of the new field.

My ask is to have a database test that inserts some bookmarks (two should be enough) with different creation/modification dates and the getbookmarks calls works correctly returning them in the expected order.

fmartingr:
> Hi @fmartingr i see all things work in PostgreSQL but when i run that show me some rollback in shiori log

Hey, I ran the test suite locally and no error showed up, where and doing what are you seeing that behavior?

fmartingr:
> 1. run new PostgreSQL database
> 2. set environment variable to use PostgreSQL `export SHIORI_DATABASE_URL=postgres://shiori:shiori@127.0.0.1:5432/shiori?sslmode=disable`
> 3. download last stable release and run that witا `./shiori server --portable`
> 4. add one random bookmark to that
> 5. clone this PR
> 6. be sure that i set environment variable with `export SHIORI_DATABASE_URL=postgres://shiori:shiori@127.0.0.1:5432/shiori?sslmode=disable` to use same database
> 7. run this PR with `go run main.go server --portable`
>    get this in PostgreSQL
> 
> ```
> error during commit: sql: transaction has already been committed or rolled back
> error during commit: sql: transaction has already been committed or rolled back
> error during commit: sql: transaction has already been committed or rolled back
> error during commit: sql: transaction has already been committed or rolled back
> ```
> 
> it is exist just in PostgreSQL not other server work correctly just see those error in log and as i test all of my migration work.

Just tested this locally and I don't see any errors on the postgresql logs.
First 1.6.3. and then this branch over the same psql database.

fmartingr:
Marked this for 1.8.0 to avoid bundling many changes in a single release.

internal/model/bookmark.go
```diff
@@ -14,7 +14,8 @@ type BookmarkDTO struct {
 	Excerpt       string `db:"excerpt"       json:"excerpt"`
 	Author        string `db:"author"        json:"author"`
 	Public        int    `db:"public"        json:"public"`
-	Modified      string `db:"modified"      json:"modified"`
+	Created       string `db:"created_at"    json:"created"`
```
fmartingr, line 17: Can we name this fields `CreatedAt` / `ModifiedAt` as well?

internal/view/content.html
```diff
@@ -38,6 +37,9 @@
 		<div id="content" dir="auto" v-pre>
 			$$.HTML$$
 		</div>
+        <div id="footer">
```
fmartingr, line 40: Can this be a `<footer>` element rather than a `div#footer` ?

internal/view/content.html
```diff
@@ -48,15 +50,26 @@
...
 			data: {
-				modified: "$$.Book.Modified$$"
+				created: "$$.Book.Created$$"
 			},
 			methods: {
-				localtime() {
-					var strTime = this.modified.replace(" ", "T");
-					if (!strTime.endsWith("Z")) {
-						strTime += "Z";
+				createdTime() {
```
fmartingr, line 56: I see that we are only using `dayjs` here. Maybe we can remove this dependency and provide a vanilla JS parseable datetime? If we return ISO strings (and we should) we can do `new Date(isoString)` and that should be a parsed datetime, we can then display that however we like. What do you think?

internal/view/assets/less/common.less
```diff
@@ -487,6 +487,64 @@ a {
...
+		padding: 20px;
+		max-width: 840px;
+		margin-bottom: 16px;
+		background-color: var(--contentBg);
+		border: 1px solid var(--border);
+		display: flex;
+		flex-flow: column;
+		align-items: center;
+
+		#metadata {
```
fmartingr, line 502: Avoid using IDs for styling, use classes instead.

internal/database/database_test.go
```diff
@@ -424,3 +427,90 @@ func testGetAccounts(t *testing.T, db DB) {
 		assert.Equal(t, 0, len(accounts))
 	})
 }
+
+func testModifiedTimeUpdate(t *testing.T, db DB) {
+	ctx := context.TODO()
```
fmartingr, line 432: Since this test uses `time.Sleep` let's run it in parallel to avoid blocking other tests.

```suggestion
func testModifiedTimeUpdate(t *testing.T, db DB) {
        t.Parallel()
	ctx := context.TODO()
```

internal/database/database_test.go
```diff
@@ -424,3 +427,90 @@ func testGetAccounts(t *testing.T, db DB) {
...
+	assert.NotEqual(t, resultBook[0].Modified, resultUpdatedBooks[0].Modified)
+	assert.Equal(t, resultBook[0].Created, resultUpdatedBooks[0].Created)
+	assert.Equal(t, resultBook[0].Created, resultBook[0].Modified)
+	assert.NoError(t, err, "Get bookmarks must not fail")
+
+	assert.Equal(t, updatedBook.Title, resultUpdatedBooks[0].Title, "Saved bookmark must have updated Title")
+}
+
+func testModifiedAndCreateTimeFilter(t *testing.T, db DB) {
+	ctx := context.TODO()
```
fmartingr, line 459: Since this test uses `time.Sleep` let's run it in parallel to avoid blocking other tests.

```suggestion
func testModifiedAndCreateTimeFilter(t *testing.T, db DB) {
	t.Parallel()
	ctx := context.TODO()
```

internal/database/database_test.go
```diff
@@ -424,3 +427,90 @@ func testGetAccounts(t *testing.T, db DB) {
 		assert.Equal(t, 0, len(accounts))
 	})
 }
+
+func testModifiedTimeUpdate(t *testing.T, db DB) {
```
fmartingr, line 431: Can you add a docstring defining exactly what we are testing there or a better function name? (`testUpdateBookmarkUpdatesModifiedTime` ?)

internal/database/database_test.go
```diff
@@ -22,6 +23,8 @@ func testDatabase(t *testing.T, dbFactory testDatabaseFactory) {
 		"testCreateBookmarkWithTag":         testCreateBookmarkWithTag,
 		"testCreateTwoDifferentBookmarks":   testCreateTwoDifferentBookmarks,
 		"testUpdateBookmark":                testUpdateBookmark,
+		"testModifiedTimeUpdate":            testModifiedTimeUpdate,
+		"testModifiedAndCreateTimeFilter":   testModifiedAndCreateTimeFilter,
```
fmartingr, line 27: `testGetBoomarksWithTimeFilters` ?

fmartingr (changes requested):
Great job @Monirzadeh !! This is starting to look real good 👏 

I have left some requests around. Most are just style things but there are some concerns around. Please take a look and let me know what you think.

internal/model/bookmark.go
fmartingr: For Golang code, `CamelCase`, for JSON `snake_case`. Though honestly we could just match everything to the same `CamelCase` in golang and avoid writting so many json annotations. What do you think?

DesarrolloAntonio: > @DesarrolloAntonio what do you think?

Regarding backward compatibility in the Shiori app, I have seen that there is an option to set one or more alternative names. I have never tried it, but theoretically, it should work.`    

```
@SerializedName("createdAt", alternate = ["created_at"])
val createdAt: String?,
```

fmartingr:
> @fmartingr as you see when i use `t.Parallel()` when two test run side by side than number and id of item in second test can be diffident so it can give use false negative and affect by each test run. for example this time it failed in Linux but other pass that i think it is not related to the platform how do you think i should solve that.

Ok I was thinking that we received a new empty database on each run but right now that's only true for SQLite (not for Postgres/Mysql). Let's remove the `t.Parallel` for now but please add a `TODO` comment:

```go
// TODO: Consider using `t.Parallel()` once we have automated database tests spawning databases using testcontainers.
```

PR #903, opened 2024-05-08
feat: deleted time added to the database

Monirzadeh:
This PR will record deleted time with bookmark id in a seprate table.
any specific unit test needed?
# Why we need this?
for create a sync functionality we need to have 3 things `created_at` `modifieded_at` and `deleted_at`
the last one added in this PR. as `deleted_at` use case is just for sync than keep that in separate table make things cleaner.

after merge #894 #896 and #903  i can finally send PR for add sync functionality to Shiori (some part of code is ready in local system). so user can sync their apps with Shiori server.


fmartingr:
I'm still not sold on this because it changes a foundational thing about Shiori and it may be confusing for users (that delete no longer means delete). I'd rather think in a better way for third-parties to know when some bookmarks are gone. Is knowing when a bookmark has been deleted the only remaining problem?

PR #905, opened 2024-05-11
ci: increase bsd test vm cpu count

fmartingr:
Increases the CPU count for the OpenBSD test to check if that influences the test total execution time.

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -122,7 +122,7 @@ jobs:
           version: ${{ matrix.os.version }}
           shell: bash
           memory: 1G
-          cpu_count: 1
+          cpu_count: 2
           run: |
             sudo pkg_add -u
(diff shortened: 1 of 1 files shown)


fmartingr:
@pacoesteban Any ideas why the ARM64 VM takes 1hr to run the test suite? I'm reluctant to spent time on this unless it's an easy fix. If that's not the case I would just remove the OpenBSD/ARM64 test (leaving the x86 in there).

pacoesteban:
I have no idea and, unfortunately, I don't have an ARM64 available to test at the moment.
I would say, leave only x86 for now.  I'll try to do some tests and get back to you if I find something.

fmartingr:
Yeah I removed it in #907, there's no rush since that test is optional, but I don't want all CI to last for 1h for each PR 😓  No need to research, just wanted to know if you knew something, thanks for answering.

PR #916, opened 2024-05-19
feat: Home button clear search query

### LINKED ISSUE #914: Home button should clear search query
**From:** https://github.com/go-shiori/shiori/discussions/871

## Acceptance criteria
- Pressing home button should remove the query from the search bar and refresh the results

Monirzadeh:
click on Home will clear search field and load home in initiate state
close #914  
Related discussion #871


fmartingr (changes requested):
I don't particularly like instancing a new Vue app to have only the $bus methods available, but since we have Vue outdated and we would need to take a look at this further in the future probably requiring a more in-depth refactor, this looks good to me now, and I like the use of events.

I've just left a comment because I think the event listener should be in a different place.

internal/view/assets/js/page/home.js
```diff
@@ -153,6 +156,12 @@ export default {
 		},
 	},
 	methods: {
+		clearHomePage() {
```
fmartingr, line 159: I think this function should only clear the homepage, the bind to the event should be made in the `mounted` method to only run that once.

fmartingr (changes requested):
Hey sorry, I think my in line comment didn't save or I didn't hit submit or something.

internal/view/assets/js/page/home.js
fmartingr: I meant that the `this.$bus.on` should be on mount (if I understand Vue correctly) so that is binded only once when the component is loaded. Unsure if there's a better method for that.

The `clearHomePage` method should only `search=""` && `this.searchBookmarks()`. That should be cleaner.

Maybe we should make it work only if the user is already at the homepage and not coming from another page? That can be done later in another PR, though.

fmartingr: This is what I was referring to, yes: 31b392ac3905ff2460c996939f427947bc994d45

fmartingr (approved):
Great work :)

PR #920, opened 2024-05-25
chore: check for avx2 processor feature when trying to run bun

Monirzadeh:
this PR fix style script to update commands if cpu not support avx2 more information
https://github.com/oven-sh/bun/issues/762#issuecomment-1186505847
it is test on linux but not on macos


scripts/styles.sh
```diff
@@ -6,14 +6,25 @@ OUTPUT_STYLECSS=internal/view/assets/css/style.css
 INPUT_ARCHIVECSS=internal/view/assets/less/archive.less
 OUTPUT_ARCHIVECSS=internal/view/assets/css/archive.css
 
+# Detect support of avx2
+
+if [ -x "$(command -v bun)" ]; then
+    if grep -q avx2 /proc/cpuinfo; then
+        BUNEXCUTE="bun"
```
Oppen, line 13: Maybe just call the variable `BUN`, the same way we don't write `CC_EXECUTE` but just `CC` to refer generically to the local C compiler.

scripts/styles.sh
```diff
@@ -6,14 +6,25 @@ OUTPUT_STYLECSS=internal/view/assets/css/style.css
...
 OUTPUT_ARCHIVECSS=internal/view/assets/css/archive.css
 
+# Detect support of avx2
+
+if [ -x "$(command -v bun)" ]; then
+    if grep -q avx2 /proc/cpuinfo; then
+        BUN="bun"
+    else
+        BUN="sde -chip-check-disable -- bun"
+        echo "Your cpu not support avx2 so We use sde for more information please lookat https://github.com/oven-sh/bun/issues/762#issuecomment-1186505847"
```
Oppen, line 16: Some minor spelling fixes:
```suggestion
        echo "Your CPU does not support avx2 so we use sde, for more information please look at https://github.com/oven-sh/bun/issues/762#issuecomment-1186505847"
```

Oppen (approved):

scripts/styles.sh
```diff
@@ -6,14 +6,25 @@ OUTPUT_STYLECSS=internal/view/assets/css/style.css
 INPUT_ARCHIVECSS=internal/view/assets/less/archive.less
 OUTPUT_ARCHIVECSS=internal/view/assets/css/archive.css
 
+# Detect support of avx2
+
+if [ -x "$(command -v bun)" ]; then
```
fmartingr, line 11: I believe it's easier to just set `BUN=bun` by default and then check support for `avx2` modifying the command:

```sh
BUN=bun
if ! rep -q avx2 /proc/cpuinfo; then
    BUN="sde -chip-check-disable -- bun"
    echo "Your CPU does not support avx2 so we use sde, for more information please look at https://github.com/oven-sh/bun/issues/762#issuecomment-1186505847"
fi
```

fmartingr (changes requested):
Hey, awesome work figuring this out. I just had some minor request.

scripts/styles.sh
fmartingr: Sadly I don't have access to a Intel Mac, but ignoring that clause (because macos does not have `/proc/cpuinfo`) works well on mac ARMs. See this patch and let me know if it works on your machine
[macos.patch](https://github.com/go-shiori/shiori/files/15446921/macos.patch)

fmartingr (approved):

PR #924, opened 2024-06-01
feat: allow selecting light/dark/follow themes in the webui

### LINKED ISSUE #251: Allow users to select between light, dark and follow system themes
Right now theme selection is inconsistent, since we can just check to use dark theme or not, but when the checkbox is not checked the browser may follow the system theme.

The idea here is to update this setting to allow selecting three options:
- Light theme
- Dark theme
- Follow system

Options should be self-explanatory

---

## Original comment

Upon installation, the webpage has the dark theme by default. Under settings, the dark theme toggle is unchecked. Toggling it and refreshing the webpage doesn't seem to enable the light theme.

Monirzadeh:
This PR try to fix #251 and some other fix
- [x] fix warrning in login page
- [x] Login page allways follow system settings theme
- [x] have 3 option Follow system settings, Light , and dark theme
- [x] hide old NightMode from settings and replace that with drop down menu
- [x] fix bug that theme not apply if you select dark mode in browser
- [x] content page follow same settings
- [x] fix content page load-settings wrong syntax that block from load settings completely  
- [x] drop down menu follow theme color
- [x] fix version number be always black in login page. now it follow the theme color

Closes #251
Closes #561


internal/view/assets/js/page/setting.js
```diff
@@ -21,9 +21,13 @@ var template = `
...
-                <input type="checkbox" v-model="appOptions.NightMode" @change="saveSetting">
-                Use dark theme
-            </label>
+                Theme &nbsp;
+                <select v-model="appOptions.Theme" @change="saveSetting">
+                <option value="follow">Follow system</option>
+                <option value="light">Light theme</option>
+                <option value="night">Dark theme</option>
+                </select>
+            </lable>
```
fmartingr, line 30:
```suggestion
            </label>
```

internal/view/assets/js/page/setting.js
```diff
@@ -21,9 +21,13 @@ var template = `
                 Hide bookmark's excerpt
             </label>
             <label>
```
fmartingr, line 23: I would put this setting the first on the list

internal/view/content.html
```diff
@@ -63,18 +63,34 @@
 						ShowId = (typeof opts.config.ShowId === "boolean") ? opts.config.ShowId : false,
 						ListMode = (typeof opts.config.ListMode === "boolean") ? opts.config.ListMode : false,
 						NightMode = (typeof opts.config.NightMode === "boolean") ? opts.config.NightMode : false,
-						UseArchive = (typeof opts.config.UseArchive === "boolean") ? opts.config.UseArchive : false;
+                        Theme = (typeof opts.config.Theme === "string" && opts.config.Theme !== "") ? opts.config.Theme : "follow",
```
fmartingr, line 66: This seems to have different whitespace type than the other lines (probably spaces instead of tabs?)

internal/view/content.html
```diff
@@ -63,18 +63,34 @@
...
 						ListMode: ListMode,
 						NightMode: NightMode,
+						Theme: Theme,
 						UseArchive: UseArchive,
 						CreateEbook: CreateEbook,
 					};
 
-					document.body.className = NightMode ? "night" : "";
+					//document.body.className = NightMode ? "night" : "";
+                    switch(Theme){
```
fmartingr, line 80: This block is repeated around. Can we refactor it into a function somehow?

internal/view/login.html
```diff
@@ -48,7 +48,7 @@
 				</div>
 			</form>
 		</div>
-		<div>
+		<div id="version">
 			<p>$$.Version$$</p>
 		</div>
```
fmartingr, line 53:
```suggestion
		<footer>
			<p>$$.Version$$</p>
		</footer>
```

internal/view/assets/less/common.less
```diff
@@ -11,17 +11,16 @@
...
-		}
-	}
+select {
+	background: var(--contentBg);
+	border: 1px solid var(--border);
+	border-radius: 4px;
+	color: var(--color);
+}
+#version {
+	color: var(--color);
```
fmartingr, line 23: Matching my suggestion below.

```suggestion
}

footer {
	color: var(--color);
```

fmartingr (changes requested):
Awesome as always! This is something I had on my radar for quite a while now since I depend a lot on my system switching themes depending on the hour of the day.

I've given this a quick look and left some comments around. Let me know what you think.

internal/view/content.html
fmartingr: `make style` only renders the less files into css, it does not modify the html/js files. What I have setup is auto formatting on save on my editor.

fmartingr:
> i don't know why but in some situation it need to refresh page.

hey I also forgot to send you this patch, I don't know how it will behave with the latest changes. It's a UI improvement to have the `select` at the right of the block:

[system-theme-justify.patch](https://github.com/user-attachments/files/15532996/system-theme-justify.patch)

Screenshot 2024-06-03 at 13 19 39

fmartingr:
Can you check if with your changes you also solved #561?

internal/view/assets/js/page/base.js
```diff
@@ -30,6 +29,7 @@ export default {
 	data() {
 		return {
 			dialog: {},
+			darkModePreference: self.matchMedia("(prefers-color-scheme: dark)"),
```
fmartingr, line 32: Is this needed?

internal/view/login.html
```diff
@@ -149,10 +148,7 @@
 					});
 				},
 				loadSetting() {
-					var opts = JSON.parse(localStorage.getItem("shiori-account")) || {},
-						NightMode = (typeof opts.config.NightMode === "boolean") ? opts.configNightMode : false;
-
-					this.NightMode = NightMode;
+                    document.body.classList.remove("light", "dark");
```
fmartingr, line 151: Does login page have any class set up? Since the user is not logged in it should just follow the system setting, right?

internal/view/content.html
```diff
@@ -62,19 +62,18 @@
 					var opts = JSON.parse(localStorage.getItem("shiori-account")) || {},
 						ShowId = (typeof opts.config.ShowId === "boolean") ? opts.config.ShowId : false,
```
fmartingr, line 63: When we refactor the frontend potentially updating to Vue 3 we need to find a better way to set up the default settings without having them set in multiple places.

fmartingr (changes requested):
Great work @Monirzadeh 👏 I have been wanting this for quite some time.

I've left some comments of code parts I think are not in use, can you confirm?

fmartingr (approved):

PR #943, opened 2024-06-25
feat: use thumbnail as ebook cover

Monirzadeh:
this PR try to test different solution for eBook cover.
- first solution is use thumbnail as cover simple and fast but aspect ration is not the best


internal/core/ebook.go
```diff
@@ -62,6 +62,11 @@ func GenerateEbook(deps *dependencies.Dependencies, req ProcessRequest, dstPath
 
 	ebook.SetTitle(book.Title)
 	ebook.SetAuthor(book.Author)
+	if deps.Domains.Storage.FileExists(bookmarkThumbnailPath) {
+		absoluteCoverPath := fp.Join(deps.Config.Storage.DataDir, bookmarkThumbnailPath)
```
fmartingr, line 66:
```suggestion
		// TODO: Use `deps.Domains.Storage` to retrieve the file.
		absoluteCoverPath := fp.Join(deps.Config.Storage.DataDir, bookmarkThumbnailPath)
```

fmartingr (approved):
This is a good first step. 👍

PR #951, opened 2024-07-03
fix: better style for quotes

### LINKED ISSUE #950: Quotes are not well rendered in archives
## Data
- **Shiori version**: 1.7.0
- **Database Engine**: Postgre
- **CLI/Web interface/Web Extension**:

### markup

<figure class="wp-block-pullquote">
<blockquote>
<p>text of the quote</p><cite>Quote author</cite></blockquote></figure>

### Quote rendered in the original article

Screenshot 2024-07-03 at 11 47 02

### Quote rendered by Shiori

Screenshot 2024-07-03 at 11 49 01

## Expected behavior
Quotes should be clearly rendered as...quotes and not like being part of the text.

## To Reproduce
1. Archive a link which has quotes
2. Open it

Monirzadeh:
This PR add better style for quotes in readable version.
close #950 

@romu70 @fmartingr any suggestion for better style? specifically in this design use Red is not too much eye caching for a quote. we can change design completely different.

**Preview**
qoute


romu70:
My take

romu70:
Fixed.

We can also think about having a soft color background instead of the gray, or having a color accent.

fmartingr:
I would use something simple, a box with a bit darker/lighter background depending on the theme, the text in italics and probably nothing else, to match the simplicity of the rest of the themes.

Once we start working on custom css users could customize this to their needs.

fmartingr:
> @fmartingr any other change?
> 
> dark light

Honestly I would be much more simple than that, I don't like extra distractions on transformations on the readable version.

Screenshot 2024-07-04 at 08 24 51

```less
blockquote {
	padding: 15px;
	margin: 15px;
	font-style: italic;
	background: var(--bgqoute);
}
```

Also I would put this style under the `#content` section, so it only applies to the readable version.

fmartingr (approved):
Thank you!

PR #956, opened 2024-07-17
Update Dockerfile.compose

cyberpunk3351:
feat: update golang


fmartingr:
Hey, thanks for the submission but we maintain ourselves on the stable channel. We probably need to unify the dockerfiles properly to use the golang version on the go.mod file, either by arguments or by installing the appropriate version directly.

PR #958, opened 2024-07-28
fix: change archive to bookmark

Monirzadeh:
as @fmartingr mention make archive publicly available can be misleading so i change that to bookmark


fmartingr (approved):
Thanks for taking care of this! 👍

PR #969, opened 2024-08-25
fix: temporary files not being deleted

### LINKED ISSUE #966: shiori update doesn't clear cache after it's finished, maxing out /tmp/ dir.
## Data
- **Shiori version**: 1.7.0
- **Database Engine**: If unknown, SQLite is the default.
- **Operating system**: Fedora GNU/Linux

I ran `shiori update 500-600` to update the descriptions, archives and images for 100 bookmarks. After it was completed, the last few hadn't worked because they 'ran out of storage'. I have 50GB free on my disk, so didn't know what it meant. Turns out it had filled up the `/tmp/` directory, which I think is RAM / cache storage. This resulted in my computer running very slowly until I restarted and it was cleared.

## Expected behavior
After a bookmark update is finished, it shouldn't need to have anything in `/tmp/` anymore and clear it.

## To Reproduce
1. Run `shiori update` with a certain amount of entries, depending on how much ram you have. e.g. it took about 100 entries to fill up my 4GB ram.
2. Notice that they may start failing and give you an error relating to not enough space available.
3. Stop the command.
4. See lots of files in `/tmp/` prefixed with either `archive` or `image` and suffixed by random numbers. Notice that the `/tmp/` directory has little or no space left in it.

Oppen:
Fixes a path mismatch between a temporary file and its removal.
It also fixes a second removal of the file.

Should fix #966


Oppen:
The reason is the `FS` domain assumes everything is relative to the data dir, while the temporary file is relative to the root directory. Why are we using the abstraction? Doesn't the os package work well enough?
Fixing it under the abstraction would require bigger changes to the domains module I believe. Not against it but I can't commit to do it myself right now.

fmartingr (approved):
As I said in #966, this should be in in order to fix the problem. i don't think we need any more changes for now in order to work with ephemeral files.

Oppen:
> it is good to have an issue or something in road map to review all the abstraction layer to fix in that level later. @Oppen can you review that when ever you have free time?

Sure!

> any way thanks for investigation and PR. as i see windows test pass for now so no problem in my side
❤️

PR #975, opened 2024-09-02
SQL Fix for mariadb-galera

beyondwatts:
Resolives mariadb-galera database "Error 1054 (42S22): Unknown column 'modified' in 'field list' (500)"


fmartingr:
Hey @beyondwatts, thank you for your contribution! Can we add a test for the `GetTags` database method to ensure this is still working for non-clustered installations/mysql?

PR #979, opened 2024-09-14
fix: bookmarks full width while in list mode

Monirzadeh:
in big screen list mode bookmarks not get full width
#### Before:
before
#### After:
after


fmartingr (approved):
👍

Issue #986, opened 2024-10-05
go.mod requires a newer version of go than what docker-compose uses


ishmaelavila:
## Data
- **Shiori version**: Main Branch / 1.7.0
- **Database Engine**: Postgres 15
- **Operating system**: Ubuntu 20

## Describe the bug / actual behavior
When attempting to build the docker containers via `docker-compose up` for the first time, everything builds and executes successfully except for Shiori itself which fails to build with the following error.
<img width="656" alt="error" src="https://github.com/user-attachments/assets/34b6f6b2-ef86-4cc1-bd1d-278a47ae3245">

## Expected behavior
Expected Shiori HTTP server to run

## To Reproduce
1. Clone a fresh version of Shiori
1. Ensure you don't have any existing docker containers/cache
1. Run `docker-compose build && docker-compose up` or similar 
1. See error

## Screenshots
Linked above

## Notes
Bumping the base layer to match what is listed in `go.mod` solves the issue. 
```
FROM docker.io/golang:1.22.3-alpine3.19
```

ishmaelavila:
https://github.com/go-shiori/shiori/pull/987
One liner to bump Go Docker Image tag

PR #987, opened 2024-10-05
chore: bump compose go version to match go.mod

ishmaelavila:
Resolves https://github.com/go-shiori/shiori/issues/986 by updating Docker Go image version to match version required in `go.mod`.


fmartingr (approved):
We missed that in our dependency upgrade, thank you for taking care of it 👍

PR #1001, opened 2024-10-19
docs: add mac launch agent guide

cdaein:
From #999 

Instructions for Mac LaunchAgent so the Shiori server will start at login and also set the environment variable to keep the login state.

Note:
I tested on my Mac but it would be nice if someone else can also do a test to make sure.


cdaein:
My bad. found some typo errors. will create another one.

PR #1002, opened 2024-10-19
docs: add Mac LaunchAgent instructions

cdaein:
From #999 

Instructions for Mac LaunchAgent so the Shiori server will start at login automatically.

Note:
I tested on my Mac but it would be nice if someone else can also do a test to make sure.
Thanks!

### DIFF
--- a/docs/Frequently-Asked-Question.md
+++ b/docs/Frequently-Asked-Question.md
@@ -236,3 +236,34 @@ There are several methods to run `shiori` on start up, however the most recommen
     ```sh
     systemctl enable --now shiori
     ```
+
+If you are using Mac, create `local.app.shiori.plist` file in `~/Library/LaunchAgents` and use the template below. Add your own secret key and paths. The filename can be anything but it's a good practice to start it with `local`:
+
+```xml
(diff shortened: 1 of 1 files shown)


docs/Frequently-Asked-Question.md
```diff
@@ -236,3 +236,34 @@ There are several methods to run `shiori` on start up, however the most recommen
     ```sh
     systemctl enable --now shiori
     ```
+
```
fmartingr, line 239: I would add it as a separate section here:

```
### How to run `shiori` on start up (macOS)?
```

And rename the other one to specify Linux

fmartingr (changes requested):
Thank you for the contribution! Really appreciated.

I think is best if we put the macOS setup under it's own section, I've left a comment about that. Just don't forget to update the table of contents and we're good to go :)

docs/Frequently-Asked-Question.md
cdaein: Made the requested changes. Please let me know if anything looks not ok. (I'm not very used to PR). Thanks.

fmartingr (approved):
This looks good! Thank you for your contribution. :)

PR #1023, opened 2024-12-09
feat: allow import of newer pocket data export files in csv format

melnary:
Recently Pocket changed the format of their data export function, switching from HTML files to a zipped folder containing CSV files, delivered in parts of up to 10000 bookmarks.
Other bookmark managers were also affected by this, as seen here: https://github.com/hoarder-app/hoarder/issues/570

This patch allows the `pocket` sub-command to handle these files correctly, expanding it's functionality to import both HTML and CSV files, as I believe the ability to import older exports may still be useful. This is done by a simple file extension check, which I think will suffice for most users.

Note that the command still does not process the multiple part files explicitly, so the user has to import them one by one, which I don't think will be too much of a hassle for the user either way. :)

I would appreciate it if someone could test this patch on an older HTML file (or give me one so I could test it myself :3), as I was only able to test it with my own CSV export.

### DIFF
--- a/internal/cmd/pocket.go
+++ b/internal/cmd/pocket.go
@@ -1,14 +1,21 @@
 package cmd
 
 import (
+	"context"
+	"encoding/csv"
+	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
+	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/PuerkitoBio/goquery"
 	"github.com/go-shiori/shiori/internal/core"
+	"github.com/go-shiori/shiori/internal/database"
 	"github.com/go-shiori/shiori/internal/model"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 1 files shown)


fmartingr:
Hey @melnary, thanks for the contribution! I'm not super famiiar with pocket and their exports so for me if you tested it it's good to go. Can you take a look at the CI errors and correct them though?

melnary:
@fmartingr Hi! I was wondering where the CI was at :3
Do we have something to address coverage in the cmd package?

fmartingr:
Dont worry about coverage, the linter errors are here: https://github.com/go-shiori/shiori/actions/runs/12276136781/job/34257042686#step:3:1

melnary:
@fmartingr Please approve the workflows again, my local linter is reporting no errors, should be good to go!

melnary:
Ah and one more thing! I did manage to track down an [older HTML export file](https://github.com/mufidu/pocket-to-markdown/blob/master/ril_export.html), to make sure that I didn't end up breaking the HTML import feature, luckily it all still works smoothly!

fmartingr:
> Ah and one more thing! I did manage to track down an [older HTML export file](https://github.com/mufidu/pocket-to-markdown/blob/master/ril_export.html), to make sure that I didn't end up breaking the HTML import feature, luckily it all still works smoothly!

That's awesome!

fmartingr (approved):
Thank you for the contribution!

Panphico:
Hi, thank you all for the patch. However, this patch use a mandatory "cursor" field (Invalid CSV format. Header must be: title,url,time_added,cursor,tags,status) that is not in my pocket extract. Could you make this field optional please ?

PR #1038, opened 2024-12-27
Update docker-compose.yaml

noticons:
fixed deprecated "links to" and changed to depends_on and setup a shiori network so each sub-package doesn't need an external network port.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -3,32 +3,32 @@
 version: "3"
 services:
   shiori:
-    build:
-      context: .
-      dockerfile: Dockerfile.compose
+    image: ghcr.io/go-shiori/shiori
     container_name: shiori
     ports:
-      - "8080:8080"
+      - "8080:8080" #set port number if it needs to be different
     volumes:
       - "./dev-data:/srv/shiori"
       - ".:/src/shiori"
     restart: unless-stopped
-    links:
+    depends_on:
       - "postgres"
       - "mariadb"
     environment:
       SHIORI_DIR: /srv/shiori
       #SHIORI_DATABASE_URL: mysql://shiori:shiori@(mariadb)/shiori?charset=utf8mb4
       SHIORI_DATABASE_URL: postgres://shiori:shiori@postgres/shiori?sslmode=disable
-
+    networks:
+      shiori:
   postgres:
     image: postgres:15
     environment:
       POSTGRES_PASSWORD: shiori
       POSTGRES_USER: shiori
     ports:
-      - "5432:5432"
-
+      - 5432
+    networks:
+      shiori:
   mariadb:
     image: mariadb:11
     environment:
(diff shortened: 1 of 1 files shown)


docker-compose.yaml
```diff
@@ -3,32 +3,32 @@
 version: "3"
 services:
   shiori:
-    build:
-      context: .
-      dockerfile: Dockerfile.compose
+    image: ghcr.io/go-shiori/shiori
```
fmartingr, line 6: Please revert this to the original state, since this docker compose is meant for development purposes.

docker-compose.yaml
```diff
@@ -3,32 +3,32 @@
...
     container_name: shiori
     ports:
-      - "8080:8080"
+      - "8080:8080" #set port number if it needs to be different
```
fmartingr, line 9: Please revert this to the original state, since this docker compose is meant for development purposes.

docker-compose.yaml
```diff
@@ -3,32 +3,32 @@
...
+      shiori:
   postgres:
     image: postgres:15
     environment:
       POSTGRES_PASSWORD: shiori
       POSTGRES_USER: shiori
     ports:
-      - "5432:5432"
-
+      - 5432
```
fmartingr, line 29: Using the short form will make ports unexposed to the local environment.

```suggestion
      - "5432:5432"
```

docker-compose.yaml
```diff
@@ -37,4 +37,10 @@ services:
       MYSQL_USER: shiori
       MYSQL_PASSWORD: shiori
     ports:
-      - "3306:3306"
+      - 3306
```
fmartingr, line 40: Using the short form will make ports unexposed to the local environment.

```suggestion
      - "3306:3306"
```

docker-compose.yaml
```diff
@@ -37,4 +37,10 @@ services:
...
+    networks:
+      shiori:
+networks:
```
fmartingr, line 43: Adding a new line for looks

```suggestion

networks:
```

docker-compose.yaml
```diff
@@ -3,32 +3,32 @@
 version: "3"
```
fmartingr, line 3: I believe we can also remove the `version` attribute since it's deprecated.

```suggestion
version: "3"
```

fmartingr (changes requested):
Hey @noticons, thank you for taking some time to update the compose file. I've added a few comments below since this docker compose is not meant to be used to deploy Shiori but only for local development, please take a look at the comments and let me know if you have any questions.

docker-compose.yaml
noticons: I understand

docker-compose.yaml
noticons: Excellent

docker-compose.yaml
noticons: My mistake! You're right.

PR #1040, opened 2024-12-28
docs(kubernetes): add tmp volume

hackeryarn:
When trying to generate ebooks in a kubernetes hosted instance, I noticed a failure due to permissions in `/tmp`. 

The best way to make sure that `/tmp` acts as expected, is to provide an `emptyDir` volume which will get cleaned up on container restart, which would be the expected behavior for a `/tmp` folder.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -93,6 +93,9 @@ spec:
       - name: app
         hostPath:
           path: /path/to/data/dir
+      - name: tmp
+        emptyDir:
+          medium: Memory
       containers:
       - name: shiori
         image: ghcr.io/go-shiori/shiori:latest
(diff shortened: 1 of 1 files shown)


fmartingr:
Thank you for fixing this part of the docs! 

I used to have Shiori under kubernetes but went back to raw docker and then to LXC, so I miss all these issues since I can't test all configurations.

hackeryarn:
@fmartingr I'll be running Shiori under kubernetes for the foreseeable future. I'll try to keep an eye out for any kubernetes related issues that pop up, but also feel free to tag me if you need help looking into any of those issues.

fmartingr:
> @fmartingr I'll be running Shiori under kubernetes for the foreseeable future. I'll try to keep an eye out for any kubernetes related issues that pop up, but also feel free to tag me if you need help looking into any of those issues.

Awesome, thank you. Likewise, let me know if there's something failing that I need to take a look into.

PR #1049, opened 2025-01-02
fix: incorrectly set cookie's expires value in login.js

Promix953:
When writing a cookie using document.cookie, the expires value should use the UTCString format.

https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie#write_a_new_cookie

If expires is set incorrectly, the cookie will expire at the end of session and thumbnail will fail to load.


fmartingr (approved):
Hey @Promix953, thank you for taking your time and fixing this issue. Right now it looks good but later on I'm going to provide the ISO directly on the login/refresh APIs to keep that kind of logic out of the frontend. Thank you!

Issue #1051, opened 2025-01-08
Support for MariaDB?


robertsilen:
Any reason not to mention MariaDB in the README? Is MariaDB supported? Cheers

fmartingr:
Closed via #1052, thanks for the contribution!

PR #1052, opened 2025-01-08
MariaDB in README.md

robertsilen:
Add MariaDB to README as suggested in #1051


fmartingr (approved):
LGTM! I should probably add tests for both databases, just in case.

PR #919, opened 2024-05-22
feat: add PWA support

ragekit:
Closes #325

PR #1060, opened 2025-02-15
feat: add PWA support share functionality

pudymody:
This is based on #919. As i said in my comment it adds the share functionality.
It only allows to install it as a native app and other apps to share to it. Offline functionality could be explored in the future.

In the home page, it checks if the query param "url" or "description" are present, and if thats the case, it automatically opens the add modal dialog with the data prefilled. From there, you can always edit it before saving.
This also allows not only the native share functionality, but to trigger this from anywhere else as you only need to construct the correct url.

`url` maps to the url to save. `description` to the excerpt and `name` to the title.
For example: `shiori.tld/?url=github.com&name=pudymody&description=pudymody+is+a+developer`

Changed the display mode from "fullscreen" to "standalone" as fullscreen hides native controls needed for basic functionality. Fullscreen is targeted more towards things like games and not apps. https://developer.mozilla.org/en-US/docs/Web/Manifest/Reference/display

Removed the 512 size icon as it didnt exists and i couldnt find some bigger version to downscale and generate it.

Used relative urls for start_url and share_target to allow subfolder installations to work without needing to define the manifest file as some templated file https://developer.mozilla.org/en-US/docs/Web/Manifest/Reference/start_url

Closes #919


fmartingr (approved):
Hey @pudymody, this looks awesome! I have tested it in my iOS emulator and it works well (except for the share target that is not supported in iOS :( ).

fmartingr:
Here are some icons in different sizes, in case you need any. PNG with transparecy.

[shiori-logo-144](https://github.com/user-attachments/assets/bfa0b464-8848-4dbe-8cf0-6f07358cbcef)
[shiori-logo-152](https://github.com/user-attachments/assets/d0d43d77-f95c-47d4-b96e-ecc262d50f49)
[shiori-logo-512](https://github.com/user-attachments/assets/f0cef242-47de-42a4-af8d-63b81436e8eb)
[shiori-logo-1024](https://github.com/user-attachments/assets/030983df-4808-4d2d-aa0e-b8813519aa80)
[shiori-logo-white-144](https://github.com/user-attachments/assets/8c2e097f-e606-4452-a21d-6582f91bc3b9)
[shiori-logo-white-152](https://github.com/user-attachments/assets/f468d68d-d380-48b9-9ab7-eb21afeb9f23)
[shiori-logo-white-512](https://github.com/user-attachments/assets/1b77c921-32bf-4813-ba76-13367462d47c)
[shiori-logo-white-1024](https://github.com/user-attachments/assets/534f1183-f16c-4190-b615-99eef249ddd6)

pudymody:
> I have one nit, could we use the same parameter names as the fields in the form? (title and excerpt) To prevent confusion.

Nice catch.

I also added the 192x192 icon variant as its the minimum recommendable. https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Guides/Making_PWAs_installable#required_manifest_members

fmartingr:
> > I have one nit, could we use the same parameter names as the fields in the form? (title and excerpt) To prevent confusion.
> 
> Nice catch.
> 
> I also added the 192x192 icon variant as its the minimum recommendable. https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Guides/Making_PWAs_installable#required_manifest_members

awesome, thank you for making the changes so quick! 

I've fixed the lint errors, hope you don't mind. Thank you for your contribution!

PR #1063, opened 2025-02-22
chore(deps): go 1.24 & alpine 3.21

fmartingr:
- Upgrade go to 1.24
- Upgrade heroku's go to 1.24
- Upgrade Alpine to 3.21


fmartingr:
Failing on openbsd due to missing go 1.24 (since it's fairly new). Let's wait on it to merge this PR.

fmartingr:
@pacoesteban Any hints around how to use openbsd 7.7 with go 1.23.3? It seems the [port exists](https://openports.pl/path/lang/go) but the system is downloading the quirks for 7.1?

pacoesteban:
> I'm not pinning into any scpecific BSD version, I was just trying to make it work. If we can have 1.24.3 running in any BSD version I'm ok with that. I guess I could just download the binary from the golang releases and that's it, right?

I guess.  If they provide binaries for OpenBSD and the Go build env is correctly set up in the VM I can't see why not.
You could also test against `-current` if the GH actions workflow you use supports it (I have no idea if that's even possible), as any ports updates are always done against `-current` and the version of shiori in stable (7.7 in this case) is already tested and won't change.  This would be a better scenario, as the new development of shiori would be tested against the future version of OpenBSD.  But this may be difficult to achieve depending on supported images and the like.

Issue #1039, opened 2024-12-28
Feedback on installation with Kubernetes and nginx reverse proxy


eric-b:
## Data
- **v1.7.2**
- **Database Engine**: postgresql.
- **Operating system**: docker image ghcr.io/go-shiori/shiori:v1.7.2 (with Kubernetes)

## Describe the bug / actual behavior

I noticed two inconsistencies in documentation related to nginx reverse proxy and Kubernetes deployment.

### Nginx reverse proxy configuration

[Documented configuration](https://github.com/go-shiori/shiori/blob/master/docs/Configuration.md#nginx) for nginx reverse proxy does not work (only change is my local service name):

```
location /shiori {
    proxy_pass http://shiori-service:8080/;
    proxy_set_header Host $host;
    proxy_set_header X-Real-IP $remote_addr;
}
```

With related configuration of container:

```
spec:
  template:
    spec:
      containers:
      - name: shiori
        image: ghcr.io/go-shiori/shiori:v1.7.2
        command: ["/usr/bin/shiori", "server", "--webroot", "/shiori"]
        ports:
        - containerPort: 8080
        env:
          - name: SHIORI_HTTP_ROOT_PATH
            value: "/shiori"
```

With this configuration, root endpoint works but any asset (css, scripts) fails with HTTP 404.

To make it work, I had to add a trailing slash to the location:

```
location /shiori/ {
    proxy_pass http://shiori-service:8080/;
    proxy_set_header Host $host;
    proxy_set_header X-Real-IP $remote_addr;
} 
```

### Security context documented for filesystem in container does not work

With [documented security context](https://github.com/go-shiori/shiori/blob/master/docs/Installation.md#using-kubernetes-manifests), archive won't work, with these logs (look for **permission denied**):

```
2024/12/28 13:00:52 File download not successful for image URL: https://techcrunch.com/wp-content/uploads/2024/12/data-breach-2024-roundup.jpg?resize=1200,675                                                                                                                                                                                             │
│ 2024/12/28 13:00:53 File download not successful for image URL: https://techcrunch.com/wp-content/uploads/2015/02/cropped-cropped-favicon-gradient.png?w=192                                                                                                                                                                                               │
[...]

PR #1065, opened 2025-02-23
docs: improve kubernetes deployment docs

hackeryarn:
This tries to improve the docs enough to sufficiently cover the ask of https://github.com/go-shiori/shiori/issues/1039.

This is the configuration that I am currently running, and it it stripped down to the simplest deployment that works with all the features.


docs/Installation.md
```diff
@@ -85,10 +85,6 @@ spec:
       labels:
         app: shiori
     spec:
-      securityContext:
```
fmartingr, line 88: I'm not running mine in k8s anymore, but last time I did this was necessary, did we fix it along the way?

fmartingr (approved):
Thank you for the update! I love how you explained every step to make it work :)

docs/Installation.md
hackeryarn: Good question on how it got fixed. I tried it in k3s and kind, just to try a more restrictive environment, and it ran without the security context in both.

PR #1102, opened 2025-04-30
error handling / URL already exists

rept0id:
This is my first pull request for this project and I'm coming ambitious. Please, show some understanding.

So, I noticed that in general, adding a bookmark with a URL that already exists, results in very weird error messages.
 - in older versions, you get something like "json failed parsing", because the endpoint panics the API and the UI doesn't expect that
 - in newer versions, the UI expects such cases and it even expects error messages ! But still, all it can show is a "Unknown error occurred" because the API instead of returning a polished error it panics.

**The whole case here is that for less tech savvy people, such errors make the project look like it's broken**, while in reality I find it to be very stable and mature. Don't catch me wrong, but a tech savvy person my one day try to add again a URL that already added maybe some months ago and "JSON/UNKNOWN ERROR" is not something this person gets.

**Let's say first, that panicking is not the best option. Specially if something is so simple as that a URL is just a duplicate.** Imagine that other people may use the same deployment and they requests may fail, just because another person made a request for a duplicate URL.
For more serious errors, panicking may be optimal instead of keeping working wrongly. This is not the case here.

Changes :

 - **Since the UI expects a polished error, I return a polished error.** (http response change)
 - **Since the UI has a kind of "modeled" way of expecting an error, I added a "ResponseHttpError" type in our model, so we can all have a common base of returning an error through our HTTP responses.** (modeling change)

Thank you for your time !

P.S.
_For "ResponseHttpError" in our model, I made a file "responses" for all this in our model, there other people can add other standard responses too._ 

_Please, I don't know how to express this in our codebase, but the responses.go in the model that I added, is not for whatever responses different endpoints return, but for some standard responses like an error response in our case._ 

_I believe and I've seen that there are other standard responses coming in the way, so I guess this will be handful for everyone._ 

_Maybe, just maybe, I should had named this "standardResponses.go" or just write a comment there, but to be fair I noticed that the rest of the files had one simple word as a name, so I decided to not over-engineer by myself and see the opinions of the rest of you ! Let's over-engineer all together at least. ;)_

### DIFF
--- a/Dockerfile.compose
+++ b/Dockerfile.compose
@@ -3,7 +3,7 @@
 # Please do not run this Dockerfile in any environment that is not
 # a local development scenario as this is not throroughly updated nor
 # tested.
-FROM docker.io/golang:1.22-alpine
+FROM docker.io/golang:1.23.6-alpine
 
 WORKDIR /src/shiori
 
--- a/internal/model/responses.go
+++ b/internal/model/responses.go
@@ -1,0 +1,5 @@
+package model
+
+type ResponseHttpError struct {
+	Error string `json:"error"`
+}
--- a/internal/webserver/handler-api.go
+++ b/internal/webserver/handler-api.go
@@ -225,6 +225,14 @@ func (h *Handler) ApiInsertBookmark(w http.ResponseWriter, r *http.Request, ps h
 	// Save bookmark to database
 	results, err := h.DB.SaveBookmarks(ctx, true, *book)
 	if err != nil || len(results) == 0 {
+		if strings.Contains(err.Error(), "UNIQUE constraint failed: bookmark.url") {
+			w.Header().Set("Content-Type", "application/json")
+			w.WriteHeader(http.StatusConflict)
+			json.NewEncoder(w).Encode(model.ResponseHttpError{Error: "URL already exists"})
+
+			return
+		}
+
 		panic(fmt.Errorf("failed to save bookmark: %v", err))
 	}
 


internal/core/url.go
```diff
@@ -36,12 +36,26 @@ func queryEncodeWithoutEmptyValues(v nurl.Values) string {
 	return buf.String()
 }
 
+// Parse parses a URL. If it fails, it tries again with "https://" prefix.
+func Parse(url string) (*nurl.URL, error) {
+	urlParsed, err := nurl.Parse(url)
+	if err != nil || urlParsed.Scheme == "" || urlParsed.Hostname() == "" {
```
fmartingr, line 42: You can parse URLs correctly without an scheme: https://go.dev/play/p/qEcMoNLqsHD

We should check for an error: `if err != nil` and after that (if no errors) if the url does not have an scheme we can just set `urlParsed.Scheme = https`: https://go.dev/play/p/8iaLlsvn1PV

internal/core/url.go
```diff
@@ -36,12 +36,26 @@ func queryEncodeWithoutEmptyValues(v nurl.Values) string {
...
+
 // RemoveUTMParams removes the UTM parameters from URL.
 func RemoveUTMParams(url string) (string, error) {
 	// Parse string URL
-	tmp, err := nurl.Parse(url)
-	if err != nil || tmp.Scheme == "" || tmp.Hostname() == "" {
-		return url, fmt.Errorf("URL is not valid")
+	tmp, err := Parse(url)
+	if err != nil {
+		return url, err
```
fmartingr, line 58: Let's make errors traceable now that we are on it

```suggestion
		return url, fmt.Errorf("error removing utm params: %w", err)
```

Dockerfile.compose
fmartingr, line 1: Please discard these changes, let's upgrade dependecies all at the same time in a different PR. There's one already open.

docker-compose.yaml
fmartingr, line 1: Please discard these changes, this docker-compose is intented for development purposes only so there's no need to make it more complex for first time contributors not familiar with containers.

internal/model/responses.go
```diff
@@ -0,0 +1,5 @@
+package model
+
+type ResponseHttpError struct {
```
fmartingr, line 3: This can be moved to `internal/http/response/error.go` and it can be called `ErrorResponse`.

internal/webserver/handler-api.go
```diff
@@ -225,6 +225,14 @@ func (h *Handler) ApiInsertBookmark(w http.ResponseWriter, r *http.Request, ps h
 	// Save bookmark to database
 	results, err := h.DB.SaveBookmarks(ctx, true, *book)
 	if err != nil || len(results) == 0 {
+		if strings.Contains(err.Error(), "UNIQUE constraint failed: bookmark.url") {
```
fmartingr, line 228: This is only temporary, so make it work with all database engines, as that error code is specific to SQLite.

- SQLite: `UNIQUE constraint failed: x`
- MySQL/MariaDB: `Duplicate entry 'x' for key 'x'`
- PostgreSQL: `duplicate key value violates unique constraint "x"`

```suggestion
                errStr := strings.ToLower(err.Error())
		if (strings.Contains(errStr, "unique constraint failed") || strings.Contains(errStr, "duplicate entry")) && 
   strings.Contains(errStr, "duplicate key value violates unique constraint") {
```

fmartingr (changes requested):
Hey @rept0id, thank you for taking care of the errors in the server. It is a really nice quality of life addition, and we will need to follow through and improve it in all areas.

Regarding your changes, let me address those separately:

A. This looks good, just pointed out to where the actual response structure need to be.
B. Put some comments around, but please discard the changes to the compose and containers, I'd like to address dependency upgrades separately and all at the same time so I can test it. Regarding the compose, since it's only for development I don't think we need to make it more complex that it is right now, unless you think there's something missing.
C. mostly good but I think we can do a little better in the error handling.

Thank you for the time you took on this! Really appreaciated.

internal/core/url.go
rept0id: done !

Dockerfile.compose
rept0id: done !

rept0id:
@fmartingr shall I do something about the failing tests ?

fmartingr:
> @fmartingr shall I do something about the failing tests ?

If you could add tests for the added/changes in `url.go`, just so we improve coverage. :)

rept0id:
updated, fixed complain of golangci "should merge variable declaration with assignment on next line (staticcheck)"

@fmartingr

PR #1105, opened 2025-05-24
Add Auth Forward method

pudymody:
Hi, i started using Authelia for my SSO and Shiori is one of the apps i use that currently doesnt support neither proxy forwarding headers or OIDC.

This is my attempt at the first as its easier to implement.

The default settings are tailored to my use case and should be enough for almost everyone, but open to different ones as this is a new world to me.

I've been testing it locally for now and everything seems to be working

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -66,7 +66,7 @@ type HttpConfig struct {
 	DisableKeepAlive             bool          `env:"HTTP_DISABLE_KEEP_ALIVE,default=true"`
 	DisablePreParseMultipartForm bool          `env:"HTTP_DISABLE_PARSE_MULTIPART_FORM,default=true"`
 
-	SSOEnable       bool     `env:"SSO_ENABLE,default=false"`
+	SSOEnabled      bool     `env:"SSO_ENABLED,default=false"`
 	SSOHeaderName   string   `env:"SSO_HEADER_NAME,default=Remote-User"`
 	SSOTrustedProxy []string `env:"SSO_TRUSTED_PROXY,default=10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16, fc00::/7"`
 }
@@ -156,7 +156,7 @@ func (c *Config) DebugConfiguration(logger *logrus.Logger) {
 	logger.Debugf(" SHIORI_HTTP_IDLE_TIMEOUT: %s", c.Http.IDLETimeout)
 	logger.Debugf(" SHIORI_HTTP_DISABLE_KEEP_ALIVE: %t", c.Http.DisableKeepAlive)
 	logger.Debugf(" SHIORI_HTTP_DISABLE_PARSE_MULTIPART_FORM: %t", c.Http.DisablePreParseMultipartForm)
-	logger.Debugf(" SHIORI_SSO_ENABLE: %t", c.Http.SSOEnable)
+	logger.Debugf(" SHIORI_SSO_ENABLED: %t", c.Http.SSOEnabled)
 	logger.Debugf(" SHIORI_SSO_HEADER_NAME: %s", c.Http.SSOHeaderName)
 	logger.Debugf(" SHIORI_SSO_TRUSTED_PROXY: %v", c.Http.SSOTrustedProxy)
 }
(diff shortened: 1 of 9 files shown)


internal/http/middleware/auth.go
```diff
@@ -46,6 +70,57 @@ func (m *AuthMiddleware) OnRequest(deps model.Dependencies, c model.WebContext)
...
+func (m *AuthMiddleware) ssoAccount(deps model.Dependencies, c model.WebContext) *model.AccountDTO {
+	if !deps.Config().Http.SSOEnabled {
+		return nil
+	}
+
+	remoteAddr := c.Request().RemoteAddr
+	ip, _, err := net.SplitHostPort(remoteAddr)
+	if err != nil {
+		var addrErr *net.AddrError
+		if errors.As(err, &addrErr) && addrErr.Err == "missing port in address" {
```
pudymody, line 82: While testing locally this always came as ip:port, as the documentation says https://pkg.go.dev/net/http#Request

But the last run on pipelines was a clear ip, i added this just in case 🤷🏻‍♂️

internal/view/index.html
```diff
@@ -160,19 +160,15 @@
 				},
 
 				onLoginSuccess() {
-					this.loadSetting();
 					this.loadAccount();
+					this.loadSetting();
```
pudymody, line 164: This swap and the following remove of check is because now we can have a valid session without a token or a previous login, and we need to have the account information already defined to prevent an exception

internal/config/config.go
```diff
@@ -65,6 +65,10 @@ type HttpConfig struct {
 	IDLETimeout                  time.Duration `env:"HTTP_IDLE_TIMEOUT,default=10s"`
 	DisableKeepAlive             bool          `env:"HTTP_DISABLE_KEEP_ALIVE,default=true"`
 	DisablePreParseMultipartForm bool          `env:"HTTP_DISABLE_PARSE_MULTIPART_FORM,default=true"`
+
+	SSOEnabled      bool     `env:"SSO_ENABLED,default=false"`
```
fmartingr, line 69: Should we rename this settings to `SSOProxyAuth...`? In case we ever support more than one SSO auth method in the future.

internal/config/config.go
```diff
@@ -65,6 +65,10 @@ type HttpConfig struct {
...
+	SSOHeaderName   string   `env:"SSO_HEADER_NAME,default=Remote-User"`
+	SSOTrustedProxy []string `env:"SSO_TRUSTED_PROXY,default=10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16, fc00::/7"`
```
fmartingr, line 71:
```suggestion
	SSOTrustedProxies []string `env:"SSO_TRUSTED_PROXY,default=10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16, fc00::/7"`
```

internal/http/middleware/auth.go
```diff
@@ -14,13 +16,35 @@ import (
 // request context.
 type AuthMiddleware struct {
 	deps model.Dependencies
+
+	trustedIPs []*net.IPNet
 }
 
 func NewAuthMiddleware(deps model.Dependencies) *AuthMiddleware {
-	return &AuthMiddleware{deps: deps}
+	plainIPs := deps.Config().Http.SSOTrustedProxy
```
fmartingr, line 24: Can we move the SSO logic to a new middleware? That way we would only inject it if needed (`SSOEnabled == true`)

Having the logic split would prove useful and is better for long term maintenance. You could even put it in a different file (`auth_sso_proxy.go`?)

internal/http/middleware/auth.go
fmartingr: Did you manage to reproduce this? (So we can add a test case)

internal/http/middleware/auth.go
```diff
@@ -46,6 +70,57 @@ func (m *AuthMiddleware) OnRequest(deps model.Dependencies, c model.WebContext)
 	return nil
 }
 
+func (m *AuthMiddleware) ssoAccount(deps model.Dependencies, c model.WebContext) *model.AccountDTO {
```
fmartingr, line 73: I would use `getSSOAccount(...) (*model.AccountDTO, error)` instead of just logging internally, so we can log all the errors directly in the parent function. What do you think?

internal/view/index.html
fmartingr: Let's dogfood this a bit in an RC before moving it forward with it. It should work in all cases from the looks of it, but I'd like to be sure.

fmartingr (changes requested):
Hey @pudymody, thanks a lot for the implementation! Everything is looking good, but I've pointed some things that I'd like in a different way. Check my comments and let me know if you agree. Thanks!

internal/http/middleware/auth.go
pudymody: I couldnt reproduce it, but i added a test with both cases

fmartingr (approved):
This is awesome! Thanks for your contribution. Hope people can integrate better with this, there are some projects around that bundle Shiori behind a layer of auth and with this they would not need to login again into Shiori 🌟

vacero:
I've just set up shiori and was looking for something like this as I set up most of my stuff behind Authelia. Waiting eagerly for the release!

Issue #1110, opened 2025-05-30
Pocket import fails with CSV parsing error


LordEidi:
## Data
- **Shiori version**: 1.7.4
- **Database Engine**: SQLite
- **Operating system**: Debian stable
- **CLI/Web interface/Web Extension**: CLI

## Describe the bug / actual behavior
When trying to import a current Pocket export, Shiori throws an error and cancels the import.

```
./shiori pocket part_000000.csv 
Invalid CSV format. Header must be: title,url,time_added,cursor,tags,status
```

When looking into the Pocket export, I noticed, that the columns have probably changed and are now like this:

`title,url,time_added,tags,status`

Looking at the import code, the lack of the column "cursor" fails the process, but the column is never used for the import anyway.

## Expected behavior
Shiori importing the list of URLs from the Pocket export.

## To Reproduce
- Get a current export from Pocket
- Extract the .csv from the ZIP file
- Get the current build from Shiori from [https://github.com/go-shiori/shiori/releases/tag/v1.7.4](https://github.com/go-shiori/shiori/releases/tag/v1.7.4)
- Start the Pocket import with 

`./shiori pocket part_000000.csv`

- You will get the mentioned error.

## Notes

I noticed, that the pocket importer requires the fields to be in a specific order. It probably would make sense to parse the title line for the expected columns and then dynamically assign the column location to the specific column import.  The columns are hardcoded on this line:

[https://github.com/go-shiori/shiori/blob/master/internal/cmd/pocket.go#L138](https://github.com/go-shiori/shiori/blob/master/internal/cmd/pocket.go#L138)

Depending on further changes on the format, this will break again when the order of the fields change.

LoisGNS:
I ran into the same issue and, as a workaround, opened the CSV in Excel and added a "cursor" column where it was needed, then re-saved. However, the resulting file had an extra (trailing) comma at the end of each row and threw another error. I used notepad++ to remove all the extra commas; also had to delete an extra blank line at the end.

I'm trying to get all my pocket URLS saved in one way or another before the deadline. While some are "real" independent URLs, and so can be available after the Pocket shutdown, others are quite old pocket-redirects, and some are on the Pocket site itself, so I've prioritized those.

rafacouto:
Try this: https://github.com/go-shiori/shiori/issues/252#issuecomment-2934962779

LoisGNS:
[...]

PR #1112, opened 2025-06-07
fix: parse pocket new CSV format

bakito:
fixes: #1110

This PR fixes Pocket CSV Parsing errors, by not verifying the whole header line, but just finding the relevant colums and using these dynamic colum indexes.

### DIFF
--- a/internal/cmd/pocket_test.go
+++ b/internal/cmd/pocket_test.go
@@ -5,7 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
-	
+
 	"github.com/go-shiori/shiori/internal/database"
 )
 
@@ -27,7 +27,7 @@ func Test_parseCsvExport_old_format(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			file, err := os.Open("../../testdata/" + tt.fileName)
 			if err != nil {
-				t.Errorf(err.Error())
+				t.Error(err.Error())
 			}
 			defer file.Close()
 			ctx := context.TODO()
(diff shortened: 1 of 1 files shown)


bakito:
I don't see the the relation to this pull request.

As far as I understand it you describe a functionality during the import.

I changed just the parsing of the csv, so it is not depending on field order or obsolete fields.


Shouldn't your be handled in a separate PR?


Please correct me if I'm wrong

fmartingr (approved):
This looks good to me. @Monirzadeh unsure what you refer to, changes look limited to parsing, so if an URL is not creating an archive/reader now it probably wasn't creating one earlier.

@bakito This change means we don't accept old CSV files from pocket?

bakito:
> This looks good to me. @Monirzadeh unsure what you refer to, changes look limited to parsing, so if an URL is not creating an archive/reader now it probably wasn't creating one earlier.
> 
> @bakito This change means we don't accept old CSV files from pocket?

Why not?
the old implementation required a predefined sequence of csv headers.
The new implementation uses the same header names as before, it just dynamically addresses the columns by the index of the column by header name instead of the previously hard coded indexes.

Again, please correct me if I'm wrong. - I still don't see your point.

fmartingr:
> Why not? the old implementation required a predefined sequence of csv headers. The new implementation uses the same header names as before, it just dynamically addresses the columns by the index of the column by header name instead of the previously hard coded indexes.
> 
> Again, please correct me if I'm wrong. - I still don't see your point.

I have zero experience with Pocket export files to be honest, I was just making sure we weren't breaking anything.

bakito:
> > Why not? the old implementation required a predefined sequence of csv headers. The new implementation uses the same header names as before, it just dynamically addresses the columns by the index of the column by header name instead of the previously hard coded indexes.
> > Again, please correct me if I'm wrong. - I still don't see your point.
> 
> I have zero experience with Pocket export files to be honest, I was just making sure we weren't breaking anything.

Me netiher, but it's just csv. If the previous implementation worked, this modification will not break it as it uses the same headers.

bakito:
@fmartingr I added some unit tests

fmartingr:
Hey @bakito can you check the CI errors before we merge this in?

bakito:
@fmartingr It should be fixed now. 
Test issues where timezone releated.

PR #1121, opened 2025-06-24
docs: fix issue with reverse proxy webroot configuration

mannykary:
It looks like the `SHIORI_HTTP_ROOT_PATH` replaced the `SHIORI_WEBROOT` environment variable. Also, I needed to add the trailing `/` to the NGINX location config (i.e. `location /shiori/`) to get the reverse proxy configuration to work.


fmartingr (approved):
Thanks for taking care of updating our docs!

PR #1134, opened 2025-08-11
add select all/deselect all/open all buttons

keltar:
Add select all/deselect all/open selected buttons for mass operations. This enables "select all, open all, delete all, done" workflow.

Opening multiple items it quirky as it requires browser permission to open multiple tabs, but browser tooltips seems quite clear on what user is required to do, and only do that once.

Any suggestion on improvements is appreciated.


keltar:
ping

PR #1178, opened 2026-01-04
[PoC] Pluggable Archiver Proof-of-Concept

honsiorovskyi:
Hello Shiori team! :wave: 

First of all, thank you for this project, I absolutely love the idea and how lightweight it is.

However, in the modern world, some things can't be ~nice~ done purely in Go and stay lightweight.

One example is archiving some heavy pages where content is rendered by JS fetching data from some API (e.g. the very Github :see_no_evil:). Thus, it seems, there's no way to avoid using some kind of web engine, rendering the page, running the JS and then capturing properly rendered DOM.

Another one would be archiving something like YT videos, where you'd need to run the whole yt-dlp stack to get the actual content.

At the same time, the approach of using such 3rd party suites has a set of drawbacks:
1. Size and complexity: if bundled with Shiori, they'd extremely blow up the project and essentially defy its purpose and the main advantage.
2. Security: in case of using a web engine for archival, we literally run some unknown 3rd party JS code in it; even given strong sandboxing mechanisms of modern web engines, it still might be a sound idea to isolate it even further by running such processing inside more isolated environments like separate VMs or even physical machines.
3. Performance: running such kind of processing on low-power hardware (like Raspberry Pi or Intel Atom) creates substantial challenges for such hardware; in this case it also might be sound to offload such processing to more power hardware and process it opportunistically in batches while keeping Shiori itself always up and running on low-power and low-consumption hardware.
4. Storage: similar to the issue above, archiving heavy content requires a lot of storage which might not be feasible to connect to the hardware where Shiori is running.

---

Of course, one could live without such archival (or use another project altogether), but it became interesting to me to check what would take to implement an elegant-ish solution to make it possible to keep Shiori small, easy and lightweight yet make it optionally extensible with external plugins to enable integrations with external archivers.

Something like this:
image

The idea is to, alongside with the internal already existing archiver, have an external one — either completely replacing the built-in one if the user opts in, or working in parallel with it (thus, enabling both local and external archival). They both would implement a very simple internal interface, and the external archiver would just be exposing that interface over some kind of a transport (probably regular REST API would be the go to approach, but there are options here).

Then the external archiver / multiplexer service (a completely separate thing, not coupled to Shiori in any way) would be running on a separate machine (or on the same machine — wherever the user chooses to deploy it) and routing the archival requests to the backends it supports (and the user decided to configure).

---

This PR very simple proof-of-concept with the external archiver that doesn't implement any specific transport yet, but just calls an external binary (it was just the easiest way to implement and test it — I still think it makes more sense to use some kind of a network transport for that instead).

_Please keep in mind that this is a PoC and I don't have intentions to merge it as is (hence, publishing it as a draft)._

However, if you like the idea and would provide some feedback on it, I'd be glad to continue working in this direction and eventually come up with a PR that would implement the external archiver part (the one inside the Shiori box) fitting the project structure, aligned with the roadmap, nicely coded, properly documented and covered with tests :crossed_fingers: :smiley: 

Hope it all makes sense. I'll also try to explain some code decisions in the inline comments in the PR itself.

Cheers!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,15 @@ run-server: generate
 run-server-v2: generate
 	GIN_MODE=$(GIN_MODE) SHIORI_DEVELOPMENT=$(SHIORI_DEVELOPMENT) SHIORI_HTTP_SERVE_WEB_UI_V2=true go run main.go server --log-level debug
 
+## Runs server for local development with external archiver
+.PHONY: run-server-ext
+run-server-ext: generate
+	@echo "Starting Shiori with external archiver..."
+	@export SHIORI_EXTERNAL_ARCHIVER_ARCHIVE_COMMAND="mkdir -p /tmp/tmp-shiori && curl -o /tmp/tmp-shiori/archive_{ID} {URL}"; \
+	export SHIORI_EXTERNAL_ARCHIVER_HAS_COMMAND="test -f /tmp/tmp-shiori/archive_{ID} && echo true || echo false"; \
+	export SHIORI_EXTERNAL_ARCHIVER_GET_COMMAND="cat /tmp/tmp-shiori/archive_{ID}"; \
+	GIN_MODE=$(GIN_MODE) SHIORI_DEVELOPMENT=$(SHIORI_DEVELOPMENT) go run main.go server --log-level debug
+
 ## Generate swagger docs
 .PHONY: swagger
 swagger:
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ run-server-ext: generate
 	@export SHIORI_EXTERNAL_ARCHIVER_ARCHIVE_COMMAND="mkdir -p /tmp/tmp-shiori && curl -o /tmp/tmp-shiori/archive_{ID} {URL}"; \
 	export SHIORI_EXTERNAL_ARCHIVER_HAS_COMMAND="test -f /tmp/tmp-shiori/archive_{ID} && echo true || echo false"; \
 	export SHIORI_EXTERNAL_ARCHIVER_GET_COMMAND="cat /tmp/tmp-shiori/archive_{ID}"; \
-	export SHIORI_EXTERNAL_ARCHIVER_DELETE_COMMAND="rm -f /tmp/tmp-shiori/archive_{ID}";
+	export SHIORI_EXTERNAL_ARCHIVER_DELETE_COMMAND="rm -f /tmp/tmp-shiori/archive_{ID}"; \
 	GIN_MODE=$(GIN_MODE) SHIORI_DEVELOPMENT=$(SHIORI_DEVELOPMENT) go run main.go server --log-level debug
 
 ## Generate swagger docs
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,7 @@ run-server-ext: generate
 	@export SHIORI_EXTERNAL_ARCHIVER_ARCHIVE_COMMAND="mkdir -p /tmp/tmp-shiori && curl -o /tmp/tmp-shiori/archive_{ID} {URL}"; \
 	export SHIORI_EXTERNAL_ARCHIVER_HAS_COMMAND="test -f /tmp/tmp-shiori/archive_{ID} && echo true || echo false"; \
 	export SHIORI_EXTERNAL_ARCHIVER_GET_COMMAND="cat /tmp/tmp-shiori/archive_{ID}"; \
+	export SHIORI_EXTERNAL_ARCHIVER_DELETE_COMMAND="rm -f /tmp/tmp-shiori/archive_{ID}";
 	GIN_MODE=$(GIN_MODE) SHIORI_DEVELOPMENT=$(SHIORI_DEVELOPMENT) go run main.go server --log-level debug
 
 ## Generate swagger docs
--- a/internal/cmd/delete.go
+++ b/internal/cmd/delete.go
@@ -70,10 +70,12 @@ func deleteHandler(cmd *cobra.Command, args []string) {
 		for _, id := range ids {
 			strID := strconv.Itoa(id)
 			imgPath := fp.Join(cfg.Storage.DataDir, "thumb", strID)
-			archivePath := fp.Join(cfg.Storage.DataDir, "archive", strID)
-
+			
 			os.Remove(imgPath)
-			os.Remove(archivePath)
+			
+			// Use ArchiverDomain to delete archive (works for both built-in and external)
+			bookmark := &model.BookmarkDTO{ID: id}
+			deps.Domains().Archiver().DeleteArchive(bookmark)
 		}
 	}
 
--- a/internal/cmd/delete.go
+++ b/internal/cmd/delete.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
+
+	"github.com/go-shiori/shiori/internal/model"
 )
 
 func deleteCmd() *cobra.Command {
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -106,7 +106,12 @@ func initShiori(ctx context.Context, cmd *cobra.Command) (*config.Config, *depen
 	dependencies := dependencies.NewDependencies(logger, db, cfg)
 	dependencies.Domains().SetAuth(domains.NewAuthDomain(dependencies))
 	dependencies.Domains().SetAccounts(domains.NewAccountsDomain(dependencies))
-	dependencies.Domains().SetArchiver(domains.NewArchiverDomain(dependencies))
+	// Initialize domains with default implementations
+	if os.Getenv("SHIORI_EXTERNAL_ARCHIVER_ARCHIVE_COMMAND") != "" {
+		dependencies.Domains().SetArchiver(domains.NewExternalArchiver(dependencies))
+	} else {
+		dependencies.Domains().SetArchiver(domains.NewBuiltInArchiver(dependencies))
+	}
 	dependencies.Domains().SetBookmarks(domains.NewBookmarksDomain(dependencies))
 	dependencies.Domains().SetStorage(domains.NewStorageDomain(dependencies, afero.NewBasePathFs(afero.NewOsFs(), cfg.Storage.DataDir)))
 	dependencies.Domains().SetTags(domains.NewTagsDomain(dependencies))
--- a/internal/core/download.go
+++ b/internal/core/download.go
@@ -4,6 +4,8 @@ import (
 	"io"
 	"net/http"
 	"time"
+
+	"github.com/go-shiori/shiori/internal/model"
 )
 
 var httpClient = &http.Client{Timeout: time.Minute}
@@ -18,7 +20,7 @@ func DownloadBookmark(url string) (io.ReadCloser, string, error) {
 	}
 
 	// Send download request
-	req.Header.Set("User-Agent", userAgent)
+	req.Header.Set("User-Agent", model.UserAgent)
 	resp, err := httpClient.Do(req)
 	if err != nil {
 		return nil, "", err
--- a/internal/core/processing.go
+++ b/internal/core/processing.go
@@ -19,7 +19,6 @@ import (
 	"github.com/disintegration/imaging"
 	"github.com/go-shiori/go-readability"
 	"github.com/go-shiori/shiori/internal/model"
-	"github.com/go-shiori/warc"
 	"github.com/pkg/errors"
 	_ "golang.org/x/image/webp"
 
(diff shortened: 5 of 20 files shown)


Makefile
```diff
@@ -74,6 +74,16 @@ run-server: generate
 run-server-v2: generate
 	GIN_MODE=$(GIN_MODE) SHIORI_DEVELOPMENT=$(SHIORI_DEVELOPMENT) SHIORI_HTTP_SERVE_WEB_UI_V2=true go run main.go server --log-level debug
 
+## Runs server for local development with external archiver
+.PHONY: run-server-ext
+run-server-ext: generate
```
honsiorovskyi, line 79: An easy way to run this PoC. It starts the normal Shiori instance but turns on the external archiver that is essentially a very dumb thing just fetching the given URL and storing it in `/tmp/tmp-shiori`.

internal/cmd/delete.go
```diff
@@ -70,10 +72,12 @@ func deleteHandler(cmd *cobra.Command, args []string) {
...
 			imgPath := fp.Join(cfg.Storage.DataDir, "thumb", strID)
-			archivePath := fp.Join(cfg.Storage.DataDir, "archive", strID)
-
+			
 			os.Remove(imgPath)
-			os.Remove(archivePath)
+			
+			// Use ArchiverDomain to delete archive (works for both built-in and external)
+			bookmark := &model.BookmarkDTO{ID: id}
+			deps.Domains().Archiver().DeleteArchive(bookmark)
```
honsiorovskyi, line 80: Here and in other places, I've changed the hardcoded archive deletion to a call to a new `DeleteArchive` method on the archiver.

I don't quite like ad-hoc `BookmarkDTO` creation here, maybe there's a better way around it, but good enough for PoC.

internal/cmd/root.go
```diff
@@ -106,7 +106,12 @@ func initShiori(ctx context.Context, cmd *cobra.Command) (*config.Config, *depen
 	dependencies := dependencies.NewDependencies(logger, db, cfg)
 	dependencies.Domains().SetAuth(domains.NewAuthDomain(dependencies))
 	dependencies.Domains().SetAccounts(domains.NewAccountsDomain(dependencies))
-	dependencies.Domains().SetArchiver(domains.NewArchiverDomain(dependencies))
+	// Initialize domains with default implementations
+	if os.Getenv("SHIORI_EXTERNAL_ARCHIVER_ARCHIVE_COMMAND") != "" {
+		dependencies.Domains().SetArchiver(domains.NewExternalArchiver(dependencies))
+	} else {
+		dependencies.Domains().SetArchiver(domains.NewBuiltInArchiver(dependencies))
+	}
```
honsiorovskyi, line 114: Extremely simple and hacky way to chose the archiver implementation. This is the most uncertain thing for me: how exactly shall the user configure the external archiver? Though env vars (= once and for all)? Through options? Per-bookmark? Shall we have only one active archiver implementation at a time, or shall we route the archival request to both internal and external one?

This seems like a topic that needs to be discussed. At the same time, I think it should be safe to do the initial implementation with the env vars (just a nicer one), and then change to something else later (if needed).

internal/core/download.go
```diff
@@ -18,7 +20,7 @@ func DownloadBookmark(url string) (io.ReadCloser, string, error) {
 	}
 
 	// Send download request
-	req.Header.Set("User-Agent", userAgent)
+	req.Header.Set("User-Agent", model.UserAgent)
```
honsiorovskyi, line 23: I moved the `userAgent` thing to the `model` package (close to the build tags) because it needs to be imported from both `core` and `domains` packages.

internal/core/processing.go
```diff
@@ -161,31 +160,11 @@ func ProcessBookmark(deps model.Dependencies, req ProcessRequest) (book model.Bo
...
-		archivalRequest := warc.ArchivalRequest{
-			URL:         book.URL,
-			Reader:      archivalInput,
-			ContentType: contentType,
-			UserAgent:   userAgent,
-			LogEnabled:  req.LogArchival,
-		}
-
-		err = warc.NewArchive(archivalRequest, tmpFile.Name())
+		err = deps.Domains().Archiver().ArchiveBookmark(&book, req.LogArchival)
```
honsiorovskyi, line 163: This is the main change: instead of calling the `warc` archiver here directly, we call a method on the `ArchiverDomain` interface. Then whichever archiver implementation is enabled will handle this request.

internal/core/processing.go
```diff
@@ -161,31 +160,11 @@ func ProcessBookmark(deps model.Dependencies, req ProcessRequest) (book model.Bo
...
-		tmpFile, err := os.CreateTemp("", "archive")
-		if err != nil {
-			return book, false, fmt.Errorf("failed to create temp archive: %v", err)
-		}
-		defer os.Remove(tmpFile.Name())
-
-		archivalRequest := warc.ArchivalRequest{
-			URL:         book.URL,
-			Reader:      archivalInput,
-			ContentType: contentType,
```
honsiorovskyi, line 173: I've currently sacrificed these two things for simplicity. If this idea makes sense at all, we can bring them back to continue passing them to the internal `warc` (or `Obelisk`) archiver.

internal/core/processing.go
```diff
@@ -161,31 +160,11 @@ func ProcessBookmark(deps model.Dependencies, req ProcessRequest) (book model.Bo
...
 		if err != nil {
 			return book, false, fmt.Errorf("failed to create archive: %v", err)
 		}
 
-		dstPath := model.GetArchivePath(&book)
-		err = deps.Domains().Storage().WriteFile(dstPath, tmpFile)
-		if err != nil {
-			return book, false, fmt.Errorf("failed move archive to destination `: %v", err)
-		}
-
```
honsiorovskyi, line 188: This whole code is just moved to the `BuiltInArchiver` implementation.

internal/domains/archiver.go
honsiorovskyi, line 1: Essentially replaced by the `archiver_builtin.go` + a single function in the tests.

internal/domains/archiver_builtin.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// Construct the destination path using the BookmarkID
+	dstPath := model.GetArchivePath(book)
+
+	err = d.deps.Domains().Storage().WriteFile(dstPath, tmpFile)
+	if err != nil {
+		return fmt.Errorf("failed to move archive to destination: %v", err)
+	}
+
+	return nil
```
honsiorovskyi, line 43: Moved here from the `core` package.

internal/domains/archiver_builtin.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (d *BuiltInArchiver) HasArchive(book *model.BookmarkDTO) bool {
+	archivePath := model.GetArchivePath(book)
+	return d.deps.Domains().Storage().FileExists(archivePath)
+}
```
honsiorovskyi, line 49: Moved here from the `BookmarksDomain` interface/implementation to keep all archive-related methods in one place.

internal/domains/archiver_builtin.go
```diff
@@ -0,0 +1,81 @@
...
+func (d *BuiltInArchiver) GetBookmarkArchive(book *model.BookmarkDTO) (model.Archive, error) {
+	archivePath := model.GetArchivePath(book)
+
+	if !d.deps.Domains().Storage().FileExists(archivePath) {
+		return nil, fmt.Errorf("archive for bookmark %d doesn't exist", book.ID)
+	}
+
+	// FIXME: This only works in local filesystem
+	return warc.Open(filepath.Join(d.deps.Config().Storage.DataDir, archivePath))
+}
```
honsiorovskyi, line 60: Basically stayed as is (moved from the original `archiver.go`).

internal/domains/archiver_builtin.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	// Delete the archive file using the storage filesystem
+	err := d.deps.Domains().Storage().FS().Remove(archivePath)
+	if err != nil {
+		return fmt.Errorf("failed to delete archive: %v", err)
+	}
+
+	return nil
+}
```
honsiorovskyi, line 75: New thing. It's been hardcoded all around the code, I centralised it here and then calling this method through `deps.Domains().Archiver()` from all the original places.

internal/domains/archiver_external.go
honsiorovskyi, line 1: Implements exactly the same `ArchiverDomain` interface but just calls a shell script ~(such secure, much safe)~ as a PoC.

This whole implementation is meant to be rewritten to actually implement the external archiver transport communicating with any compatible implementation over the wire. This is essentially the very new thing being added to the code base, the rest is just refactoring to make it possible to use this thing.

internal/domains/bookmarks.go
```diff
@@ -17,11 +17,6 @@ func (d *BookmarksDomain) HasEbook(b *model.BookmarkDTO) bool {
 	return d.deps.Domains().Storage().FileExists(ebookPath)
 }
 
-func (d *BookmarksDomain) HasArchive(b *model.BookmarkDTO) bool {
-	archivePath := model.GetArchivePath(b)
-	return d.deps.Domains().Storage().FileExists(archivePath)
-}
-
```
honsiorovskyi, line 24: Moved to `ArchiverDomain`.

internal/domains/bookmarks.go
```diff
@@ -39,7 +34,7 @@ func (d *BookmarksDomain) GetBookmark(ctx context.Context, id model.DBID) (*mode
 
 	// Check if it has ebook and archive.
 	bookmark.HasEbook = d.HasEbook(&bookmark)
-	bookmark.HasArchive = d.HasArchive(&bookmark)
+	bookmark.HasArchive = d.deps.Domains().Archiver().HasArchive(&bookmark)
```
honsiorovskyi, line 37: Here and in other places: calling the currently running implementation of the archiver to do the corresponding operation.

internal/http/handlers/bookmark_test.go
```diff
@@ -101,6 +103,30 @@ func TestBookmarkContentHandler(t *testing.T) {
...
+	result, isFatalErr, err := core.ProcessBookmark(deps, processRequest)
+	content.Close()
+
+	if err != nil && isFatalErr {
+		return nil, fmt.Errorf("failed to process: %v", err)
+	}
+
+	return &result, nil
+}
+
```
honsiorovskyi, line 129: Moved from the original `archiver.go` / `ArchiverDomain`. There this function was called `DownloadBookmarkArchive` and this test was the only usage of it. I don't know where it would be a good place to put that code into — because the `ArchiverDomain` thing is actually doing a different thing now: instead of a _caller_ of the `core` code it became its _callee_ — but maybe it can stay here.

internal/model/archiver.go
honsiorovskyi, line 1: This is another interesting thing. Since the bookmark renderer API expects a `warc` archive to be returned and there's no way/sense to return it from the external archvier, I've created a new `Archive` interface that follows the `warc.Archive` exactly + a simple `SingleFileArchive` that just wraps whatever the PoC of the external archiver is returning into that interface.

For the real implementation of the external archiver we probably would need something more sophisticated that would actually a) support multiple resources/files, b) work over the wire to fetch them from the external archiver.

internal/model/domains.go
```diff
@@ -38,8 +36,10 @@ type AccountsDomain interface {
 }
 
 type ArchiverDomain interface {
-	DownloadBookmarkArchive(book BookmarkDTO) (*BookmarkDTO, error)
-	GetBookmarkArchive(book *BookmarkDTO) (*warc.Archive, error)
+	ArchiveBookmark(book *BookmarkDTO, logEnabled bool) error
+	GetBookmarkArchive(book *BookmarkDTO) (Archive, error)
+	HasArchive(book *BookmarkDTO) bool
+	DeleteArchive(book *BookmarkDTO) error
 }
```
honsiorovskyi, line 43: The actual archiver API. This is its internal form, and the external archiver module would implement exactly the same thing (+ maybe some more methods for granular resource retrieval — see the `Archive` interface above), but in a form of a REST API/gRPC/etc.

What I've noticed so far is that `HasArchive` is called quite often the most called method, so we'd need to pay close attention to it performance — it should be as fast/cached as we can make it not to waste resources. The rest can be relatively slow compared to it.

internal/model/main.go
```diff
@@ -10,4 +10,10 @@ var (
 const (
 	// ShioriNamespace
 	ShioriURLNamespace = "https://github.com/go-shiori/shiori"
+
+)
+
+var (
+	// UserAgent: default user agent used for downloads and archival
+	UserAgent = "Shiori/" + BuildVersion + " (" + BuildCommit + ") (+https://github.com/go-shiori/shiori)"
```
honsiorovskyi, line 18: Moved here from the `core` package.

internal/webserver/handler-api-ext.go
```diff
@@ -139,13 +139,14 @@ func (h *Handler) ApiDeleteViaExtension(w http.ResponseWriter, r *http.Request,
...
 		strID := strconv.Itoa(book.ID)
 		imgPath := fp.Join(h.DataDir, "thumb", strID)
-		archivePath := fp.Join(h.DataDir, "archive", strID)
-
+		
 		os.Remove(imgPath)
-		os.Remove(archivePath)
+		
+		// Use ArchiverDomain to delete archive (works for both built-in and external)
+		h.dependencies.Domains().Archiver().DeleteArchive(&book)
```
honsiorovskyi, line 149: Another one of those places where archive deletion was hardcoded, but now calls the archiver interface.

internal/webserver/handler-api.go
```diff
@@ -95,14 +95,13 @@ func (h *Handler) ApiGetBookmarks(w http.ResponseWriter, r *http.Request, ps htt
...
 		imgPath := fp.Join(h.DataDir, "thumb", strID)
-		archivePath := fp.Join(h.DataDir, "archive", strID)
 		ebookPath := fp.Join(h.DataDir, "ebook", strID+".epub")
 
 		if FileExists(imgPath) {
 			bookmarks[i].ImageURL = path.Join(h.RootPath, "bookmark", strID, "thumb")
 		}
 
-		if FileExists(archivePath) {
+		if h.dependencies.Domains().Archiver().HasArchive(&bookmarks[i]) {
```
honsiorovskyi, line 104: One of those many calls to `HasArchive` that we need to make fast :)

